PR #302, opened 2026-03-18
feat: RAG service API key file management

tsivaprasad:
## Summary
This PR adds `RAGServiceKeysResource`, a resource that stores provider API keys (OpenAI, Anthropic, etc.) as files on the host filesystem and mounts them into the RAG container in read-only mode. The keys are not exposed via Docker Swarm configs or API responses.

## Changes

- Add `RAGServiceKeysResource` (`swarm.rag_service_keys`) — manages the `{DataDir}/services/{instanceID}/keys/ `directory: Create writes key files at mode 0600 in a 0700 directory, Update rewrites them for key rotation, Delete removes the directory via os.RemoveAll
- Add `extractRAGAPIKeys` helper that extracts API keys from a parsed `RAGServiceConfig` into a `{pipeline_name}_embedding.key / {pipeline_name}_rag.key `filename map
- Extend `ServiceContainerSpecOptions` with `KeysPath`; when non-empty a read-only bind mount to `/app/keys` is added to the container spec
- Wire `RAGServiceKeysResource` into `generateRAGInstanceResources` after `RAGServiceUserRole`; RAG config is now parsed at resource-generation time
- Register the new resource type in `RegisterResourceTypes`

## Testing
1. Covered unit tests
2. Manual Verification

- Created cluster
- Created DB using below config:
```
{
  "id": "storefront",
  "spec": {
    "database_users": [
      {
        "username": "admin",
        "password": "password",
        "db_owner": true,
      }
    ],
    "port": 0,
    "nodes": [
    ],
    "services": [
      {
        "service_id": "rag",
        "service_type": "rag",
        "version": "latest",
        "host_ids": ["host-1"],
        "port": 0,
        "config": {
          "defaults": {
            "token_budget": 2000,
            "top_n": 10
          },
          "pipelines": [
            {
              "name": "default",
              "description": "Main RAG pipeline",
              "tables": [
                {
                  "table": "documents_content_chunks",
                  "text_column": "content",
                  "vector_column": "embedding"
                }
              ],
              "embedding_llm": {
                "provider": "openai",
                "model": "text-embedding-3-small",
                "api_key": "sk-proj-AAA"
              },
              "rag_llm": {
                "provider": "anthropic",
                "model": "claude-sonnet-4-5",
                "api_key": "sk-ant-XXX"
              },
              "token_budget": 4000,
              "top_n": 15,
              "search": {
                "hybrid_enabled": true,
                "vector_weight": 0.7
              }
            }
          ]
        }
      }
    ]
  }
}
```

- No `api_keys` in response
- Confirmed that the key files exists at path 
```
# pwd
/host-1/services/storefront-rag-host-1/keys
# ls -la
total 8
drwx------ 4 root root 128 Mar 17 16:48 .
drwx------ 3 root root  96 Mar 17 16:48 ..
-rw------- 1 root root 164 Mar 17 16:48 default_embedding.key
-rw------- 1 root root 108 Mar 17 16:48 default_rag.key

```
## Checklist

- [x]  Tests added 

## Notes for Reviewers

- RAGServiceKeysResource owns the keys/ subdirectory entirely (no separate DirResource), since no other resource shares it
- The KeysPath field on ServiceContainerSpecOptions is currently unused by MCP (zero value skips the mount) — no behaviour change for existing MCP services
- Key files are written as root-owned (UID 0); the RAG container UID and ownership will be addressed when container deployment is wired up in the next PR

[PLAT-490](https://pgedge.atlassian.net/browse/PLAT-490)

[PLAT-490]: https://pgedge.atlassian.net/browse/PLAT-490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tsivaprasad:
[@coderabbitai](https://github.com/coderabbitai) review

server/internal/orchestrator/swarm/rag_service_keys_resource.go
```diff
@@ -0,0 +1,225 @@
...
+	return filepath.Join(parentPath, "keys"), nil
+}
+
+func (r *RAGServiceKeysResource) Refresh(ctx context.Context, rc *resource.Context) error {
+	if rc == nil {
+		return resource.ErrNotFound
+	}
+	keysDir, err := r.keysDir(rc)
+	if err != nil {
+		return resource.ErrNotFound
```
jason-lynch: I think this would be better as:

```suggestion
		return err
```

`DirResourceFullPath` already returns `ErrNotFound` when appropriate and we don't want to mask other types of errors.

server/internal/orchestrator/swarm/rag_service_keys_resource.go
```diff
@@ -0,0 +1,225 @@
...
+
+func (r *RAGServiceKeysResource) Create(ctx context.Context, rc *resource.Context) error {
+	keysDir, err := r.keysDir(rc)
+	if err != nil {
+		return err
+	}
+	if err := os.MkdirAll(keysDir, 0o700); err != nil {
+		return fmt.Errorf("failed to create keys directory: %w", err)
+	}
+	if err := os.Chmod(keysDir, 0o700); err != nil {
```
jason-lynch: We're already specifying `0o700` to `MkdirAll`, so this `Chmod` is unnecessary. Did you mean to do `Chown` here? That would make sense since the RAG server probably doesn't need to run as `root`.

server/internal/orchestrator/swarm/orchestrator.go
```diff
@@ -578,41 +578,42 @@ func (o *Orchestrator) generateMCPInstanceResources(spec *database.ServiceInstan
 	orchestratorResources = append(orchestratorResources, serviceInstanceSpec, serviceInstance)
 
 	// Append per-node ServiceUserRole resources for each additional database node.
-	// The canonical resources (above) cover the first node; nodes [1:] each get
+	// The canonical resources (above) cover spec.NodeName; all other nodes get
 	// their own RO and RW role that sources credentials from the canonical.
-	if len(spec.DatabaseNodes) > 1 {
-		for _, nodeInst := range spec.DatabaseNodes[1:] {
-			perNodeRWID := ServiceUserRolePerNodeIdentifier(spec.ServiceSpec.ServiceID, ServiceUserRoleRW, nodeInst.NodeName)
+	for _, nodeInst := range spec.DatabaseNodes {
```
jason-lynch: Did you mean to include these changes? If so, could you please request a review from Ryan as well? The way we manage these users is pretty tricky, and I want to make sure we don't accidentally break another service.

server/internal/orchestrator/swarm/rag_service_keys_resource.go
```diff
@@ -0,0 +1,225 @@
...
+func (r *RAGServiceKeysResource) writeKeyFiles(keysDir string) error {
+	for name, key := range r.Keys {
+		if err := validateKeyFilename(name); err != nil {
+			return err
+		}
+		path := filepath.Join(keysDir, name)
+		if err := os.WriteFile(path, []byte(key), 0o600); err != nil {
+			return fmt.Errorf("failed to write key file %q: %w", name, err)
+		}
+		if err := os.Chmod(path, 0o600); err != nil {
```
jason-lynch: Similar to above, we're already setting `0o600` in `WriteFile`, so this `Chmod` is unnecessary. Did you mean to do `Chown`?

jason-lynch (commented):
Awesome! Thank you for addressing the other comments about the key file lifecycle. Most of my questions are pretty minor, but I do want to be really careful about the database user stuff.

rshoemaker (approved):
Looks good.

PR #329, opened 2026-04-06
feat: wire PostgREST into the orchestrator

moizpgedge:
- Route target_session_attrs=read-write for postgrest in plan_update
- PostgREST container env contains only PGRST_* vars; connection details live in postgrest.conf db-uri
- Register "latest" image version for postgrest
- populateCredentials reads from _rw role for postgrest (authenticator role)
- Guard DataDirID usage so non-data-dir services don't break
- Fix ServiceInstanceName to hash variable inputs into a 16-char base36 suffix, staying well under Docker Swarm's 63-char limit
- Add PostgRESTPreflightResource dependency on PostgresDatabaseResourceIdentifier
- Add docs/Supported-services/Postgrest-docs.md
- Add e2e/postgrest_test.go covering provisioning, lifecycle, JWT, failover, and role verification

## Summary
Wires PostgREST into the orchestrator as a fully functional service type: connection routing, container spec, credential role selection, service name collision fix, preflight dependency, docs, and E2E tests.
## Changes
- `plan_update.go`: route `target_session_attrs=read-write` for postgrest so libpq always connects to the primary
- `service_spec.go`: PostgREST container env is exactly 3 `PGRST_*` vars — connection details live in `postgrest.conf` db-uri only
- `service_images.go`: register `"latest"` image version for postgrest
- `service_instance_spec.go`: `populateCredentials` reads from `_rw` role for postgrest (authenticator); guard `DataDirID` so services without a data dir don't break
- `orchestrator.go`: fix `ServiceInstanceName` — hash all variable inputs into a 16-char base36 suffix to stay under Docker Swarm's 63-char limit
- `postgrest_preflight_resource.go`: add `PostgresDatabaseResourceIdentifier` as a dependency so preflight runs after the DB is ready
- `docs/Supported-services/Postgrest-docs.md`: operator runbook — DBA grants, config reference, deploy/remove commands
- `e2e/postgrest_test.go`: 11 E2E tests covering provisioning, JWT auth, preflight failures, role verification, config updates, multi-host, and failover

- ...

## Testing
```
# Unit tests
go test ./server/internal/orchestrator/swarm/... -v -run "PostgREST|Postgrest|ServiceInstanceName"
go test ./server/internal/workflows/... -v -run TestResolveTargetSessionAttrs

# E2E tests
go test ./e2e/... -v -run TestPostgREST

```
## Checklist

- [x] Documentation updated (if needed)
- [x] Issue is linked (branch name or URL in PR description)
- [ ] Breaking changes (if any) are clearly called out in the PR description

### DIFF
--- a/server/internal/orchestrator/swarm/orchestrator.go
+++ b/server/internal/orchestrator/swarm/orchestrator.go
@@ -3,7 +3,7 @@ package swarm
 import (
 	"bytes"
 	"context"
-	"crypto/sha256"
+	"crypto/sha1"
 	"errors"
 	"fmt"
 	"io"
@@ -162,11 +162,11 @@ func (o *Orchestrator) GenerateInstanceResources(spec *database.InstanceSpec, sc
 }
 
 // ServiceInstanceName generates a Docker Swarm service name for a service instance.
-// All three inputs are hashed together so truncated prefixes never collide across
-// different databases or services on the same host. serviceType is omitted because
-// serviceID is already unique within a database.
+// The hostID is hashed to produce a stable suffix, matching the scheme used by
+// InstanceIDFor. serviceType is omitted because serviceID is already unique within
+// a database.
 func ServiceInstanceName(databaseID, serviceID, hostID string) string {
-	hash := sha256.Sum256([]byte(databaseID + ":" + serviceID + ":" + hostID))
+	hash := sha1.Sum([]byte(hostID))
 	base36 := new(big.Int).SetBytes(hash[:]).Text(36)
 
 	// Docker Swarm service names are limited to 63 characters.
--- a/server/internal/orchestrator/swarm/orchestrator.go
+++ b/server/internal/orchestrator/swarm/orchestrator.go
@@ -168,24 +168,6 @@ func (o *Orchestrator) GenerateInstanceResources(spec *database.InstanceSpec, sc
 func ServiceInstanceName(databaseID, serviceID, hostID string) string {
 	hash := sha1.Sum([]byte(hostID))
 	base36 := new(big.Int).SetBytes(hash[:]).Text(36)
-
-	// Docker Swarm service names are limited to 63 characters.
-	// Two separators + 8-char hash = 10 fixed chars; 53 chars remain for the IDs.
-	// Give databaseID up to half (26), then serviceID gets the rest.
-	const budget = 53
-	if len(databaseID)+len(serviceID) > budget {
-		dbCap := budget / 2 // 26 max for databaseID
-		if len(databaseID) < dbCap {
-			dbCap = len(databaseID)
-		}
-		databaseID = databaseID[:dbCap]
-		svcCap := budget - dbCap
-		if svcCap > len(serviceID) {
-			svcCap = len(serviceID)
-		}
-		serviceID = serviceID[:svcCap]
-	}
-
 	return fmt.Sprintf("%s-%s-%s", databaseID, serviceID, base36[:8])
 }
 
--- a/server/internal/orchestrator/swarm/orchestrator.go
+++ b/server/internal/orchestrator/swarm/orchestrator.go
@@ -3,7 +3,7 @@ package swarm
 import (
 	"bytes"
 	"context"
-	"crypto/sha1"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"io"
@@ -162,13 +162,31 @@ func (o *Orchestrator) GenerateInstanceResources(spec *database.InstanceSpec, sc
 }
 
 // ServiceInstanceName generates a Docker Swarm service name for a service instance.
-// It follows the same host ID hashing convention used for Postgres instance IDs
-// (see database.InstanceIDFor), producing shorter, more readable names when host
-// IDs are UUIDs.
-func ServiceInstanceName(serviceType, databaseID, serviceID, hostID string) string {
-	hash := sha1.Sum([]byte(hostID))
+// All three inputs are hashed together so truncated prefixes never collide across
+// different databases or services on the same host. serviceType is omitted because
+// serviceID is already unique within a database.
+func ServiceInstanceName(databaseID, serviceID, hostID string) string {
+	hash := sha256.Sum256([]byte(databaseID + ":" + serviceID + ":" + hostID))
 	base36 := new(big.Int).SetBytes(hash[:]).Text(36)
-	return fmt.Sprintf("%s-%s-%s-%s", serviceType, databaseID, serviceID, base36[:8])
+
+	// Docker Swarm service names are limited to 63 characters.
+	// Two separators + 8-char hash = 10 fixed chars; 53 chars remain for the IDs.
+	// Give databaseID up to half (26), then serviceID gets the rest.
+	const budget = 53
+	if len(databaseID)+len(serviceID) > budget {
+		dbCap := budget / 2 // 26 max for databaseID
+		if len(databaseID) < dbCap {
+			dbCap = len(databaseID)
+		}
+		databaseID = databaseID[:dbCap]
+		svcCap := budget - dbCap
+		if svcCap > len(serviceID) {
+			svcCap = len(serviceID)
+		}
+		serviceID = serviceID[:svcCap]
+	}
+
+	return fmt.Sprintf("%s-%s-%s", databaseID, serviceID, base36[:8])
 }
 
 func (o *Orchestrator) instanceResources(spec *database.InstanceSpec, scripts database.Scripts) (*database.InstanceResource, []resource.Resource, error) {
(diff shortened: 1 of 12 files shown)


moizpgedge:
@coderabbitai review

server/internal/orchestrator/swarm/orchestrator.go
```diff
@@ -157,13 +157,20 @@ func (o *Orchestrator) GenerateInstanceResources(spec *database.InstanceSpec) (*
...
 	base36 := new(big.Int).SetBytes(hash[:]).Text(36)
-	return fmt.Sprintf("%s-%s-%s-%s", serviceType, databaseID, serviceID, base36[:8])
+	// Pad to 16 chars (sha256 produces enough entropy; left-pad with zeros if short).
+	suffix := fmt.Sprintf("%016s", base36)[:16]
+	prefix := serviceType
+	if len(prefix) > 10 {
+		prefix = prefix[:10]
+	}
+	return fmt.Sprintf("%s-%s", prefix, suffix)
 }
```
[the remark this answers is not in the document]
moizpgedge: Not valid

server/internal/orchestrator/swarm/orchestrator.go
```diff
@@ -162,13 +157,20 @@ func (o *Orchestrator) GenerateInstanceResources(spec *database.InstanceSpec) (*
...
 
 // ServiceInstanceName generates a Docker Swarm service name for a service instance.
-// It follows the same host ID hashing convention used for Postgres instance IDs
-// (see database.InstanceIDFor), producing shorter, more readable names when host
-// IDs are UUIDs.
+// Docker Swarm requires service names to be 63 characters or fewer. Because
+// databaseID and serviceID are often UUIDs or long strings, all four components
+// are hashed together into a single 16-char base-36 suffix. The serviceType
+// prefix is kept for readability and is truncated to 10 chars to fit comfortably.
 func ServiceInstanceName(serviceType, databaseID, serviceID, hostID string) string {
```
jason-lynch: What's the reason for this change? This might break existing services, and it deviates from the database instance service name implementation.

moizpgedge: This handles the case where databaseID and serviceID are UUIDs Since services are discovered by label not name, the rename just triggers a brief restart on next reconciliation rather than orphaning anything. Happy to revert if you prefer keeping it closer to the existing pattern.

jason-lynch: Yes, please revert this.

Host IDs and Database IDs need to be globally unique, whereas service IDs only need to be unique within a database, so it's much less likely that someone will use a UUID for a service ID. The previous implementation was human-readable and matched the same scheme as database instances, which is very handy when you're trying to troubleshoot.

jason-lynch: If you're looking to shorten these IDs: do we need to include `serviceType`?

The reason that the Postgres service includes the word `postgres` is that it allows for other database services that do not have a unique identifier. We don't have the same problem with services because they already have a unique identifier.

server/internal/orchestrator/swarm/orchestrator.go
```diff
@@ -162,13 +162,31 @@ func (o *Orchestrator) GenerateInstanceResources(spec *database.InstanceSpec) (*
...
-// It follows the same host ID hashing convention used for Postgres instance IDs
-// (see database.InstanceIDFor), producing shorter, more readable names when host
-// IDs are UUIDs.
-func ServiceInstanceName(serviceType, databaseID, serviceID, hostID string) string {
-	hash := sha1.Sum([]byte(hostID))
+// All three inputs are hashed together so truncated prefixes never collide across
+// different databases or services on the same host. serviceType is omitted because
+// serviceID is already unique within a database.
+func ServiceInstanceName(databaseID, serviceID, hostID string) string {
+	hash := sha256.Sum256([]byte(databaseID + ":" + serviceID + ":" + hostID))
```
jason-lynch: Could you please change this back to:

```go
	hash := sha1.Sum([]byte(hostID))
```

That will make this hash match the scheme we use for instances. Sometimes security tools alert about using sha1 because they want to make sure you're not using it for cryptographic purposes. This is not a cryptographic use case, so it's fine to ignore the alert.

jason-lynch: My main issue with changing this scheme is that it makes services and database instances different. I understand you're addressing the service name length issue. Could you please open a dedicated ticket for it and not address it in this PR? That way, we can ensure service names remain consistent across both services and database instances.

moizpgedge: Ok i will open a new ticket for this issue.

server/internal/orchestrator/swarm/orchestrator.go
moizpgedge: i have reverted this If you're looking to shorten these IDs: do we need to include serviceType?
the answer is :No

jason-lynch:
@tsivaprasad Can you please review this whenever you get a chance?

server/internal/orchestrator/swarm/orchestrator.go
```diff
@@ -162,13 +162,13 @@ func (o *Orchestrator) GenerateInstanceResources(spec *database.InstanceSpec, sc
...
-// IDs are UUIDs.
-func ServiceInstanceName(serviceType, databaseID, serviceID, hostID string) string {
+// The hostID is hashed to produce a stable suffix, matching the scheme used by
+// InstanceIDFor. serviceType is omitted because serviceID is already unique within
+// a database.
+func ServiceInstanceName(databaseID, serviceID, hostID string) string {
 	hash := sha1.Sum([]byte(hostID))
 	base36 := new(big.Int).SetBytes(hash[:]).Text(36)
-	return fmt.Sprintf("%s-%s-%s-%s", serviceType, databaseID, serviceID, base36[:8])
+	return fmt.Sprintf("%s-%s-%s", databaseID, serviceID, base36[:8])
```
tsivaprasad, line 171: Is there a specific reason for removing the service type from the service name? Pls revert this

server/internal/orchestrator/swarm/service_spec.go
tsivaprasad: As per postgrest.md file ADMIN Port is 8081 right?

server/internal/orchestrator/swarm/orchestrator.go
moizpgedge: Jason raised this in a previous comment and confirmed serviceType is not needed here. Another ticket is opened for this.

https://github.com/pgEdge/control-plane/pull/329#discussion_r3080319612

server/internal/orchestrator/swarm/service_spec.go
moizpgedge: you're right` PGRST_ADMIN_SERVER_PORT=3001` was wrong. Fixed to 8081

server/internal/orchestrator/swarm/orchestrator.go
tsivaprasad: @jason-lynch If each server (MCP/RAG/PostgREST) has around 5 services, it becomes very difficult to debug when the services have identical or similar names, as it’s hard to differentiate between them. what do you think to add service_type back to name?

tsivaprasad (approved):
Verified & Changes looks good.

1. Created cluster
2. Created JWT using script
```
TOKEN=$(python3 - <<'EOF'
    import hmac, hashlib, base64, json, time

    secret = b"a-secret-key-of-at-least-32-characters"

    def b64url(data):
        if isinstance(data, str):
            data = data.encode()
        return base64.urlsafe_b64encode(data).rstrip(b"=").decode()

    header  = b64url(json.dumps({"alg":"HS256","typ":"JWT"}))
    payload = b64url(json.dumps({"role":"app","exp":int(time.time())+3600}))
    sig     = b64url(hmac.new(secret, f"{header}.{payload}".encode(), hashlib.sha256).digest())
    print(f"{header}.{payload}.{sig}")
    EOF
    )

```

3. Created DB and service suing below
```
 curl -X POST http://localhost:3000/v1/databases \
        -H 'Content-Type: application/json' \
        --data '{
            "id": "storefront",
            "spec": {
                "nodes": [
                ],
                "services": [
                    {
                        "service_id": "api",
                        "service_type": "postgrest",
                        "version": "latest",
                        "host_ids": ["host-1"],
                        "port": 3100,
                        "config": {
                            "jwt_secret": "a-secret-key-of-at-least-32-characters"
                        }
                    }
                ]
            }
        }'
{"task":{"scope":"database","entity_id":"storefront","database_id":"storefront","task_id":"019d975d-8c2f-7b7b-acd0-8001f9eb03a4","created_at":"2026-04-16T17:36:23Z","type":"create","status":"pending"},"database":{"id":"storefront","created_at":"2026-04-16T17:36:23Z","updated_at":"2026-04-16T17:36:23Z","state":"creating","spec":{"database_name":"storefront","postgres_version":"18.3","spock_version":"5","nodes":[{"name":"n1","host_ids":["host-1"]}],"services":[{"service_id":"api","service_type":"postgrest","version":"latest","host_ids":["host-1"],"port":3100}]}}}
```
4. Verified & confirmed 
```

curl -X POST http://localhost:3100/products \
    -H "Content-Type: application/json" \
    -H "Authorization: Bearer $TOKEN" \
    --data '{"name": "Widget", "price": 9.99}'
{"code":"PGRST205","details":null,"hint":null,"message":"Could not find the table 'public.products' in the schema cache"}
```

tsivaprasad:
@moizpgedge Pls check conflicts otherwise all good and approved this PR.

PR #337, opened 2026-04-09
fix: add allowlist validation for RAG pipeline names

tsivaprasad:
## Summary
This PR adds allowlist validation for RAG pipeline names, rejecting names that don't match `^[a-z0-9_-]+$` at the API layer before any resources are provisioned. 

## Changes
- Add `ragPipelineNamePattern` regex (`^[a-z0-9_-]+$`) to `server/internal/database/rag_service_config.go`
- Validate pipeline names against the allowlist in `validateRAGPipeline`, producing a clear error message on rejection
- Add `TestParseRAGServiceConfig_PipelineNameAllowlist` covering 7 valid and 7 invalid name variants (spaces, uppercase, slashes, path traversal, dots, Unicode)

## Testing
1. Create cluster 
2. Create DB with invalid pipeline name

```
restish control-plane-local-1 create-database '{
    "spec": {
      "database_name": "test_invalid_name",
      "database_users": [{"username":"admin","password":"pw","db_owner":true,"attributes":["LOGIN","SUPERUSER"]}],
      "port": 0,
      "nodes": [{"name":"n1","host_ids":["host-1"]}],
      "services": [{
        "service_id": "rag", "service_type": "rag", "version": "latest",
        "host_ids": ["host-1"], "port": 0,
        "config": {
          "pipelines": [{
            "name": "My Pipeline",
            "tables": [{"table":"docs","text_column":"content","vector_column":"embedding"}],
            "embedding_llm": {"provider":"openai","model":"text-embedding-3-small","api_key":"sk-test"},
            "rag_llm": {"provider":"anthropic","model":"claude-haiku-4-5-20251001","api_key":"sk-ant-test"}
          }]
        }
      }]
    }
  }' | jq .

{
  "message": "services[0].config: pipelines[0].name \"My Pipeline\" is invalid: must match ^[a-z0-9_-]+$",
  "name": "invalid_input"
}

```

## Checklist

- [x] Tests added 

[PLAT-536](https://pgedge.atlassian.net/browse/PLAT-536)

[PLAT-536]: https://pgedge.atlassian.net/browse/PLAT-536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/internal/database/rag_service_config.go
+++ b/server/internal/database/rag_service_config.go
@@ -13,11 +13,12 @@ import (
 // ragPipelineNamePatternText is the allowlist pattern for RAG pipeline names.
 // It is kept as a const so that the compiled regexp and the error message both
 // reference the same literal and cannot drift apart.
-const ragPipelineNamePatternText = `^[a-z0-9_-]+$`
+const ragPipelineNamePatternText = `^[a-z0-9_][a-z0-9_-]*$`
 
 // ragPipelineNamePattern restricts pipeline names to lowercase alphanumeric
-// characters, hyphens, and underscores. This keeps key filenames
-// ({name}_embedding.key / {name}_rag.key) safe and auditable.
+// characters, hyphens, and underscores. The first character must not be a
+// hyphen so that names are safe as filename components and cannot be
+// misinterpreted as CLI flags if ever passed to a command.
 var ragPipelineNamePattern = regexp.MustCompile(ragPipelineNamePatternText)
 
 // RAGPipelineLLMConfig represents LLM configuration for an embedding or RAG step.
--- a/server/internal/database/rag_service_config.go
+++ b/server/internal/database/rag_service_config.go
@@ -10,10 +10,15 @@ import (
 	"strings"
 )
 
+// ragPipelineNamePatternText is the allowlist pattern for RAG pipeline names.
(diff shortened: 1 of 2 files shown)


server/internal/database/rag_service_config.go
```diff
@@ -4,11 +4,22 @@ import (
...
+	"regexp"
 	"slices"
 	"sort"
 	"strings"
 )
 
+// ragPipelineNamePatternText is the allowlist pattern for RAG pipeline names.
+// It is kept as a const so that the compiled regexp and the error message both
+// reference the same literal and cannot drift apart.
+const ragPipelineNamePatternText = `^[a-z0-9_-]+$`
```
jason-lynch: One thing to be careful about is that filenames that start with a `-` can be interpreted as a flag/option if they're used as a command line argument. We should make sure that they're not used that way, like `exec.Command("some_command", pipelineName)`. You could also restrict the starting character, like `^[a-z0-9_][a-z0-9_-]*$`

jason-lynch (approved):
Nice! I do have one suggestion, but it's non-blocking if you're certain that these filenames won't be used as a command line argument.

PR #338, opened 2026-04-10
feat: add connect_as field for service database credentials

rshoemaker:
## Summary                                                                                                                                                                                                                             
                                                                                                            
  - Services now specify which `database_users` entry to connect as via the new required `connect_as` field on `ServiceSpec`, replacing the auto-created `svc_*_ro`/`svc_*_rw` service accounts for MCP                                  
  - Validates `connect_as` references an existing `database_users` entry, and that `allow_writes: true` requires `db_owner: true`
  - MCP config resource reads credentials from the spec instead of `ServiceUserRole` state                                                                                                                                               
  - `ServiceUserRole` creation is disconnected for MCP (PostgREST/RAG still use it until they adopt `connect_as`)                                                                                                                        
  - State migration (v2.0.0) removes `swarm.service_user_role` resources and dependency references from existing deployments                                                                                                             
  - E2E fixtures updated with `connect_as`                                                                                                                                                                                                                                                                                                               
   
## Test plan                                                                                                 

  - Unit tests pass (976 tests)
  - Manual testing: RW mode (connect_as: "app", allow_writes: true) — reads and writes succeed
  - Manual testing: RO mode (connect_as: "app_read_only", allow_writes: false) — reads succeed, writes rejected with read-only transaction                                                                                               
  - Manual testing: config swap via update-database without container restart (SIGHUP reload)                                                                                                                                            
  - Manual testing: pg_read_all_data covers spock schema, admin-created tables, other-user tables, custom schemas                                                                                                                        
  - Golden tests regenerated                                                                                                                                                                                                             
  - State migration tested (removes service_user_role resources + dependency refs)                                                                                                                                                       
  - E2E tests against Lima fixture                                                                                                                                                                                                       
  - Verify PostgREST/RAG provisioning is unaffected


server/internal/orchestrator/swarm/service_instance_spec.go
```diff
@@ -82,6 +90,12 @@ func (s *ServiceInstanceSpecResource) TypeDependencies() []resource.Type {
 }
 
 func (s *ServiceInstanceSpecResource) populateCredentials(rc *resource.Context) error {
+	// MCP uses connect_as credentials from the spec — no ServiceUserRole lookup needed.
+	if s.ServiceSpec.ServiceType == "mcp" {
+		return nil
+	}
```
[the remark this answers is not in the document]
rshoemaker: Credentials is never set in plan_update.go — it's nil on the ServiceInstanceSpec. And populateCredentials is called during Refresh/Create/Update, which means it runs on a resource that was deserialized from etcd state. On first create, Credentials is nil. On subsequent reconciliations, the resource is loaded from state which would have whatever Credentials was last serialized.

However, for MCP, populateCredentials was never populating Credentials before either (it returns early). So Credentials for MCP resources would be nil in state. The only way it could have stale data is if this was previously a non-MCP service that got converted to MCP, which isn't a supported operation.

This finding is not a real concern. Credentials starts nil for MCP and stays nil. No fix needed.

server/internal/workflows/plan_update.go
```diff
@@ -123,6 +123,20 @@ func (w *Workflows) getServiceResources(
...
+		for _, u := range spec.DatabaseUsers {
+			if u.Username == serviceSpec.ConnectAs {
+				connectAsUser = u
+				break
+			}
+		}
+		if connectAsUser == nil {
+			return nil, fmt.Errorf("connect_as user %q not found in database_users", serviceSpec.ConnectAs)
+		}
+	}
```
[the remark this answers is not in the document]
rshoemaker: This is not worth fixing here. The password field on database_users is already validated at the API layer — it's required on create, and the API rejects empty passwords. By the time we reach the workflow planner, the password has already been validated and the user has been created in Postgres with it.

Adding a redundant empty-password check in the planner would be defensive coding against an impossible state — the API won't let you get here with an empty password. If anything breaks upstream and an empty password leaks through, the service would fail to authenticate, which is a clear and diagnosable error. No fix needed.

server/internal/orchestrator/swarm/mcp_config_resource.go
```diff
@@ -44,23 +44,19 @@ type MCPConfigResource struct {
...
 	return []string{
-		// Credentials are populated from ServiceUserRole resources during refresh.
-		"/ro_username",
-		"/ro_password",
-		"/rw_username",
-		"/rw_password",
+		// Credentials are resolved from database_users via connect_as.
+		"/connect_as_username",
+		"/connect_as_password",
 	}
```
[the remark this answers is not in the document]
rshoemaker: fixed.

server/internal/api/apiv1/validate.go
```diff
@@ -357,6 +357,9 @@ func validateServiceSpec(svc *api.ServiceSpec, path []string, isUpdate bool, nod
 		errs = append(errs, validateDatabaseConnection(svc.DatabaseConnection, dcPath, nn)...)
 	}
 
+	// Validate connect_as references a valid database_users entry
+	errs = append(errs, validateConnectAs(svc, dbUsers, path)...)
```
tsivaprasad: For now I think it should be MCP specific right?

rshoemaker: fixed.

api/apiv1/design/database.go
```diff
@@ -224,8 +224,14 @@ var ServiceSpec = g.Type("ServiceSpec", func() {
...
 	})
+	g.Attribute("connect_as", g.String, func() {
+		g.Description("Username of the database_users entry this service connects as. " +
+			"The user must exist in database_users and have appropriate roles for the service's needs.")
+		g.Example("app")
+		g.Meta("struct:tag:json", "connect_as")
+	})
 
-	g.Required("service_id", "service_type", "version", "host_ids")
+	g.Required("service_id", "service_type", "version", "host_ids", "connect_as")
```
tsivaprasad: `connect_as` should not be in Required() yet

Since RAG and PostgREST don't use connect_as yet, this will break all RAG and PostgREST create/update requests with a 400 before they reach validateServiceSpec.

Suggested fix:


// Remove "connect_as" from Required()
g.Required("service_id", "service_type", "version", "host_ids")
And either add g.Default("") to keep the generated type as string, or accept the *string it generates when optional and update validateConnectAs and convert.go accordingly. 

Otherwise all good

tsivaprasad (approved):
Looks Good

api/apiv1/design/database.go
rshoemaker: Fixed.  I used `g.Default("")` to keep the changes from rippling through other parts of the code.

PR #339, opened 2026-04-13
feat: user-defined sql scripts

jason-lynch:
## Summary

Adds the ability to run arbitrary SQL at two different points in the database creation process:

- `post_init`: This script runs immediately after the database is created, before any users are created.
- `post_database_create`: This script runs immediately after Spock is initialized, before any subscriptions have been created.

This feature can be used to perform first-time setup for roles and similar operations.

## Changes

- Added a `scripts` field to the database spec:

```js
{
  // ...
  "scripts": {
    "post_init": [
      "CREATE ROLE read_only NOLOGIN",
      "...",
      "..."
    ],
    "post_database_create": [
      "ALTER DEFAULT PRIVILEGES FOR ROLE admin GRANT USAGE ON SCHEMAS TO read_only",
      "...",
      "..."
    ]
  }
}
```

## Testing

Basic usage:

```sh
{
  "id": "storefront",
  "spec": {
    "database_users": [
      {
        "username": "admin",
        "password": "password",
        "db_owner": true,
      },
      {
        "username": "app",
        "password": "password",
        "roles": ["foo_role"]
      }
    ],
    "port": 0,
    "patroni_port": 0,
    "nodes": [
    ],
    "scripts": {
      "post_init": [
        "CREATE ROLE foo_role NOLOGIN"
      ],
      "post_database_create": [
        "CREATE TABLE foo (id int primary key, val text)",
        "INSERT INTO foo (id, val) VALUES (1, 'foo')"
      ]
    }
  }
}
EOF

# validate that admin has foo_role role
cp-psql -U admin -i storefront-n1-689qacsi -- -c "SELECT oid, rolname FROM pg_roles WHERE pg_has_role('admin', oid, 'member') and rolname = 'foo_role'"
cp-psql -U admin -i storefront-n2-9ptayhma -- -c "SELECT oid, rolname FROM pg_roles WHERE pg_has_role('admin', oid, 'member') and rolname = 'foo_role'"

# validate that foo table is populated
cp-psql -U admin -i storefront-n1-689qacsi -- -c "SELECT val FROM foo WHERE id = 1"
cp-psql -U admin -i storefront-n2-9ptayhma -- -c "SELECT val FROM foo WHERE id = 1"

# validate that replication works for table foo
cp-psql -U admin -i storefront-n1-689qacsi -- -c "INSERT INTO foo (id, val) VALUES (2, 'bar')"
cp-psql -U admin -i storefront-n2-9ptayhma -- -c "INSERT INTO foo (id, val) VALUES (3, 'baz')."
cp-psql -U admin -i storefront-n2-9ptayhma -- -c "SELECT val FROM foo WHERE id = 2"
cp-psql -U admin -i storefront-n1-689qacsi -- -c "SELECT val FROM foo WHERE id = 3"
```

After that, you can try changing your statements and validating that the changed statements do not run.

You can also try specifying an invalid statement in `post_init` for a new database, like `CREATE DATABASE foo`, and observe that the database creation process fails with an error about how `CREATE DATABASE` is not allowed within a transaction. Then remove or replace the bad statement, submit an `update-database` request, and verify that the creation process resumes where it left off and executes your updated statement.

You can also look at the included examples in the API reference, docs, and tests for other things to try.

## Notes for Reviewers

Roles added in the `pre_init` field will not carry over to new nodes. This will be addressed in a subsequent PR.

PLAT-543

### DIFF
--- a/server/internal/database/script.go
+++ b/server/internal/database/script.go
@@ -1,0 +1,192 @@
+package database
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/pgEdge/control-plane/server/internal/resource"
+	"github.com/rs/zerolog"
+	"github.com/samber/do"
+)
+
+type ScriptName string
+
+func (s ScriptName) String() string {
+	return string(s)
+}
+
+const (
+	ScriptNamePostInit           ScriptName = "post_init"
+	ScriptNamePostDatabaseCreate ScriptName = "post_database_create"
+)
+
+type Script struct {
+	DatabaseID string     `json:"database_id"`
+	NodeName   string     `json:"node_name"`
+	Name       ScriptName `json:"name"`
+	Statements []string   `json:"statements"`
+	Succeeded  bool       `json:"succeeded"`
+	// NeedsToRun is used to trigger a resource diff. Resources should call
+	// SetScriptNeedsToRun to set it.
+	NeedsToRun bool `json:"needs_to_run"`
+}
+
+type Scripts map[ScriptName]*Script
+
+type ScriptResult struct {
+	DatabaseID  string     `json:"database_id"`
+	ScriptName  ScriptName `json:"script_name"`
+	NodeName    string     `json:"node_name"`
+	Succeeded   bool       `json:"succeeded"`
+	StartedAt   time.Time  `json:"started_at"`
+	CompletedAt time.Time  `json:"completed_at"`
+	Error       string     `json:"error"`
+}
+
+func NewScriptResult(databaseID string, scriptName ScriptName, nodeName string) *ScriptResult {
+	return &ScriptResult{
+		DatabaseID: databaseID,
+		ScriptName: scriptName,
+		NodeName:   nodeName,
+	}
+}
+
+func (s *ScriptResult) Validate() error {
+	if s == nil {
+		return errors.New("cannot be nil")
+	}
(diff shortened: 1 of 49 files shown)


jason-lynch:
@coderabbitai review

server/internal/database/script.go
```diff
@@ -0,0 +1,192 @@
...
+	return nil
+}
+
+func ExecuteScript(ctx context.Context, rc *resource.Context, conn *pgx.Conn, script *Script) error {
+	ok, err := checkScriptPreconditions(rc, script)
+	if err != nil {
+		return err
+	}
+	if !ok {
+		return nil
```
rshoemaker, line 80: Thoughts on explicitly logging when scripts are NOT executed?  I think this can only happen during update-database, which might be valuable to see in the logs.  I'm not sure.

rshoemaker (approved):

server/internal/database/script.go
jason-lynch: Is there a specific use case that you're thinking of when those logs would be helpful?

As it is now, these are executed only once per node/database during the database's lifetime. They can execute during `update-database` if you're using `update-database` to retry a `create-database` that failed before the scripts executed successfully. If we were to log on non-executions, you'd see "not executing scripts" on every update or restore operation after that point.

rshoemaker: If there are too many then the message wouldn't be useful.  I was just thinking that people who don't read the docs might be confused why the script isn't running.

PR #347, opened 2026-04-16
fix: missing restore config in in-place restore

jason-lynch:
## Summary

The restore config on the database/instance spec is used to bootstrap a new node. It is not used during an in-place restore, where the active restore configuration is provided as a separate argument.

This change takes that into account and propagates that argument through to the `HasRestoreConfig` property on the database resource. Now that we're setting this property correctly, the "post restore" logic triggers as expected.

## Testing

```sh
# the TestS3BackupRestore test fails on main, but passes on this branch
make test-e2e E2E_FIXTURE=lima E2E_RUN=TestS3BackupRestore
```

PLAT-564


rshoemaker (approved):
Looks good.

PR #348, opened 2026-04-17
feat: transfer roles to new nodes

jason-lynch:
## Summary

Spock's Z0DAN feature transfers database-level objects to the new node, such as schemas, tables, and extensions. Roles are global, so they are not covered by this process.

This commit adds a mechanism to the add/populate node process to dump roles from the source node with `pg_dumpall -r` and then load them on the new node after the primary instance is created.

With this change, roles created outside of our `database_users` field, such as in a `post_init` script or in an ad-hoc `CREATE ROLE` statement, will be transferred to the new node.

## Changes

- Adds `DumpRolesResource` to dump roles from the source node in each add node operation
- Adds `RolesSourceResource` to point the instance resource to the roles dump
- Modifies the `InstanceResource` to check if a `RolesSourceResource` exists and, if so, create roles from the dump rather than the spec's `database_users` field
- Modifies `UpdateDatabase` in the operations package to consolidate the `if len(adds) > 0` blocks and to create the new resources
- Resolves the `TODO`s in the scripts E2E test

## Testing

```sh
# the scripts e2e test includes assertions for roles
make test-e2e E2E_RUN=TestScripts

# ad-hoc roles created outside of scripts will also transfer
# create a database
{
  "id": "storefront",
  "spec": {
    "database_users": [
      {
        "username": "admin",
        "password": "password",
        "db_owner": true,
      }
    ],
    "port": 0,
    "patroni_port": 0,
    "nodes": [
    ]
  }
}
EOF

# create a role in an ad-hoc SQL statement
cp-psql -U admin -i storefront-n1-689qacsi -- -c "CREATE ROLE foo WITH LOGIN"
cp-psql -U admin -i storefront-n1-689qacsi -- -c "GRANT CONNECT ON DATABASE storefront TO foo"

# add a node to the database
cp1-req update-database storefront <<EOF | cp-follow-task
{
  "id": "storefront",
  "spec": {
    "database_users": [
      {
        "username": "admin",
        "password": "password",
        "db_owner": true,
      }
    ],
    "port": 0,
    "patroni_port": 0,
    "nodes": [
    ]
  }
}
EOF

# validate that the foo role exists on the new node
cp-psql -U admin -i storefront-n2-9ptayhma -- -c "SELECT EXISTS (SELECT 1 FROM pg_catalog.pg_roles WHERE rolname = 'foo')"
# validate that foo's connect privilege has transferred
cp-psql -U foo -i storefront-n2-9ptayhma -- -c "SELECT NOW()"
```

## Notes for Reviewers

PLAT-544


server/internal/database/dump_roles.go
```diff
@@ -0,0 +1,126 @@
...
+	roles, statements, err := SanitizeRolesDump(dump.String())
+	if err != nil {
+		return fmt.Errorf("failed to sanitize pg_dumpall output: %w", err)
+	}
+
+	r.Roles = roles
+	r.Statements = statements
+
+	return nil
+}
```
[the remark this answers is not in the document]
jason-lynch: This is intentional. This code path is only used when a new node is added and populated by Spock, not for delta/in-place restores.

server/internal/database/instance_resource.go
```diff
@@ -288,3 +267,65 @@ func (r *InstanceResource) updateConnectionInfo(ctx context.Context, rc *resourc
...
+		return fmt.Errorf("failed to check for roles source: %w", err)
+	}
+
+	dump, err := resource.FromContext[*DumpRolesResource](rc, DumpRolesResourceIdentifier(source.SourceNodeName))
+	if err != nil {
+		return fmt.Errorf("failed to find roles dump from source node '%s': %w", source.SourceNodeName, err)
+	}
+
+	return r.createRolesFromDump(ctx, dump, tx)
+}
```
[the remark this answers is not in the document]
jason-lynch: This is incorrect. There is a set order to operations that's determined by the `operations.UpdateDatabase` function: updates, followed by adds, followed by deletes. If a user modifies `DatabaseUsers` in the same request as adding a new node, the update is guaranteed to run on the source node before the new node is added. So, that change is guaranteed to be in the roles dump.

server/internal/database/instance_resource.go
```diff
@@ -288,3 +267,65 @@ func (r *InstanceResource) updateConnectionInfo(ctx context.Context, rc *resourc
...
+	}
+	for i, statement := range dump.Statements {
+		_, err := tx.Exec(ctx, statement)
+		if err != nil {
+			return fmt.Errorf("failed to execute statement[%d] from roles dump: %w, %s", i, err, statement)
+		}
+	}
+
+	return nil
+}
```
jason-lynch, line 306: I will fix this.

server/internal/database/operations/update_database.go
```diff
@@ -174,3 +136,94 @@ func updateFunc(options UpdateDatabaseOptions) (func([]*NodeResources) ([]*resou
...
+				n.SourceNode = defaultSource
+			}
+
+			sourceNames.Add(n.SourceNode)
+			sourceNodeMap[n.NodeName] = n.SourceNode
+		}
+		if err := validateSourceNodes(newNames, sourceNames); err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
jason-lynch: I will fix this.

server/internal/database/dump_roles.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (r *DumpRolesResource) Delete(ctx context.Context, rc *resource.Context) error {
+	return nil
+}
+
+var sqlStatementPattern = regexp.MustCompile(`(?i)^[a-z].+;$`)
+var createRoleRegex = regexp.MustCompile(`(?i)^create role ([^;]+);`)
+
+func SanitizeRolesDump(in string) ([]string, []string, error) {
```
rshoemaker: unused error.

server/internal/database/dump_roles.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	var dump strings.Builder
+	err = orch.ExecuteInstanceCommand(ctx, &dump, primary.Spec.DatabaseID, primary.Spec.InstanceID, cmd...)
+	if err != nil {
+		return fmt.Errorf("failed to execute pg_dumpall: %w, output: %s", err, dump.String())
+	}
+
+	roles, statements, err := SanitizeRolesDump(dump.String())
+	if err != nil {
+		return fmt.Errorf("failed to sanitize pg_dumpall output: %w", err)
```
rshoemaker: unreachable - err is always nil.

server/internal/database/dump_roles.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return err
+	}
+
+	primary, err := GetPrimaryInstance(ctx, rc, r.NodeName)
+	if err != nil {
+		return fmt.Errorf("failed to get primary instance: %w", err)
+	}
+
+	cmd := []string{
```
rshoemaker, line 73: I think `pg_dumpall` outputs password hashes, which aren't being redacted in the error below: `dump.String()`

server/internal/database/instance_resource.go
```diff
@@ -288,3 +271,65 @@ func (r *InstanceResource) updateConnectionInfo(ctx context.Context, rc *resourc
...
+	for i, role := range dump.Roles {
+		err := postgres.CreateRoleIfNotExists(role).Exec(ctx, tx)
+		if err != nil {
+			return fmt.Errorf("failed to create role[%d] from roles dump: %w", i, err)
+		}
+	}
+	for i, statement := range dump.Statements {
+		_, err := tx.Exec(ctx, statement)
+		if err != nil {
+			return fmt.Errorf("failed to execute statement[%d] from roles dump: %w, %s", i, err, statement)
```
rshoemaker: Same comment here - I think `dump` contains password hashes that shouldn't be logged.

rshoemaker (dismissed):
Couple minor questions/issues.

server/internal/database/instance_resource.go
jason-lynch: Yes, this was leftover from debugging. It's already fixed.

server/internal/database/dump_roles.go
jason-lynch: Thanks. Fixed.

server/internal/database/dump_roles.go
jason-lynch: Thanks! Fixed.

server/internal/database/dump_roles.go
jason-lynch: Thanks! Fixed.

server/internal/database/dump_roles.go
rshoemaker: Sorry, I don't see the fix.  `dump.String()` might contain password hashes.  It looks like the exec is merging stdout&stderr, so maybe it's better to just wrap the error:

`return fmt.Errorf("failed to execute pg_dumpall: %w", err)`

jason-lynch: Sorry, I thought this was another repeat of the other issue that you and CodeRabbit commented on. I'll do something for this.

jason-lynch: I just went with your solution. The problem is that without the output, you don't have a real error message, just something like `exit code: 1`. We want the combined stderr output for pgbackrest, so it's not straightforward to decouple these in `ExecuteInstanceCommand`. It's unlikely that the `pg_dumpall` would exit non-zero _and_ output dump content, but you're right that it's not really worth the risk.

rshoemaker: yeah, neither option is great.  What do you think about something like:

```
  var passwordRedactRegex = regexp.MustCompile(`(?i)PASSWORD\s+'[^']*'`)                                                                                                                                                                 
                                                                                                                                                                                                                                         
  // replace line 82                                                                                                                                                                                                                     
  if err != nil {                                                                                                                                                              
      redacted := passwordRedactRegex.ReplaceAllString(dump.String(), "PASSWORD '[REDACTED]'")                                                                                                                                           
      return fmt.Errorf("failed to execute pg_dumpall: %w, output: %s", err, redacted)                                                                                                                                                   
  }                                                                                                                                                                                                                                      
```

It's your call...I'll approve the PR though.

rshoemaker (approved):

PR #354, opened 2026-04-20
fix: ability to change ports with systemd

jason-lynch:
## Summary

Fixes a bug where changing the Patroni or Postgres port would cause the update operation to fail on systemd. This happened because, in the update method, we were setting the connection info based on the updated spec.

This change means we copy the connection info from the previous instance state to retrieve the previously active Patroni port. Then, it adds a restart operation to handle cases where the Postgres port has changed.

## Changes

- Get connection info from previous instance state in `InstanceResource.Update` with a fallback to the current spec (the previous behavior) when the previous state is malformed.
- Restart Postgres if needed in `InstanceResource`
  - We need this to change the port on systemd. On Swarm, the container was already restarted at this point due to changing the port binding.
  - Note that this is a necessary change to our earlier decision to leave it up to the user to restart after updating. I think it's a net improvement to the user experience.

## Testing

The `TestPortChange` E2E exercises this fix:

```sh
make test-e2e E2E_RUN=TestPortChange
```

## Notes for Reviewers

PLAT-548

### DIFF
--- a/e2e/port_change_test.go
+++ b/e2e/port_change_test.go
@@ -1,0 +1,130 @@
+//go:build e2e_test
+
+package e2e
+
+import (
+	"testing"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/stretchr/testify/require"
+
+	api "github.com/pgEdge/control-plane/api/apiv1/gen/control_plane"
+)
+
+func TestPortChange(t *testing.T) {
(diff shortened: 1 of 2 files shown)


jason-lynch:
@coderabbitai review

rshoemaker (approved):

PR #352, opened 2026-04-17
feat: migrate PostgREST to connect_as credential model

moizpgedge:
Replace auto-created `svc_{service_id}_ro/rw` service user roles for PostgREST with the `connect_as` field — a reference to a `database_users` entry that the customer owns and manages. This aligns PostgREST with the credential model already used by MCP.

- `validateConnectAs` was not being called for `postgrest` service type — fixed: all three service types (mcp, postgrest, rag) now validate `connect_as` at the API layer.
- Added `noDefaultConnectAs` flag in validate tests to cover the missing-`connect_as` and unknown-user error paths for PostgREST.

- `PostgRESTAuthenticatorResource`: removed `UserRoleID` field, added `ConnectAsUsername`; `Dependencies()` now points to `PostgresDatabaseResource` instead of `ServiceUserRole`; SQL targets the `connect_as` user directly instead of a generated username.
- `PostgRESTConfigResource`: receives `ConnectAsUsername` / `ConnectAsPassword` at construction time from the connect_as database user instead of reading from `ServiceUserRole`.
- `generateMCPInstanceResources()`: `ServiceUserRole` generation is now skipped for both `mcp` and `postgrest`; only `rag` still creates `svc_*` roles. Per-node `PostgRESTAuthenticatorResource` instances carry `ConnectAsUsername` directly.

- `ServiceInstanceResource.Dependencies()`: removed stale PostgREST `ServiceUserRole` deps that blocked container start on the wrong resource.
- `ServiceInstanceSpecResource.Dependencies()`: removed PostgREST `ServiceUserRole` deps; added `PostgRESTPreflightResourceIdentifier` — this ensures the container does not start until the database exists, fixing a race condition where PostgREST failed with "database does not exist" when Patroni had not finished bootstrapping.
- `ServiceInstanceSpecResource.populateCredentials()`: simplified — all current service types use config-file credentials, so `s.Credentials` is always nil.

  only 5xx responses use `Error`. Eliminates spurious ERR flood from 409 conflict responses during cluster initialization.

- `docs/services/index.md`: replaced `svc_*_ro/rw` credential section with `connect_as` model description.
- `docs/services/managing.md`: updated `service_id` field description, added `connect_as` row to spec table, added `database_users` + `connect_as` to PostgREST example, updated removal warning.
- `docs/services/postgrest.md`: updated overview, all three examples now include `database_users` and `connect_as: "app"`, updated JWT role-claim note to reference `connect_as` user.

[...]

PR #351, opened 2026-04-17
fix: add replication slot resources in migration

jason-lynch:
## Summary

The replication slot resource wasn't added until v0.7.0, so databases created in versions <= v0.6.2 will be missing this resource. With this change, the migration now adds replication slot resources if they're missing in addition to upgrading any existing resources.

## Changes

- Regenerated `server/internal/resource/migrations/schemas/v1_0_0/database.go` by running:

```sh
go run -C ./server/internal/resource/migrations/schematool . -repo $(pwd) -package v1_0_0 main server/internal/database \
    InstanceResource \
    LagTrackerCommitTimestampResource \
    PostgresDatabaseResource \
    ReplicationSlotAdvanceFromCTSResource \
    ReplicationSlotCreateResource \
    ReplicationSlotResource \
    SubscriptionResource \
    SyncEventResource \
    WaitForSyncEventResource \
    > ./server/internal/resource/migrations/schemas/v1_0_0/database.go
```

- Updated the state v1.0.0 migration to incorporate the new `DatabaseID` field on `PostgresDatabaseResource`
- Updated the state v1.0.0 migration to add replication slot resources when they don't exist

## Testing

```sh
# This patch re-adds the old IP address setting to make this template work with
# v0.6.2.
git apply <<EOF
diff --git a/e2e/fixtures/roles/deploy_control_plane/templates/stack.yaml.tmpl b/e2e/fixtures/roles/deploy_control_plane/templates/stack.yaml.tmpl
index 05cb104b..47c88d75 100644
--- a/e2e/fixtures/roles/deploy_control_plane/templates/stack.yaml.tmpl
+++ b/e2e/fixtures/roles/deploy_control_plane/templates/stack.yaml.tmpl
@@ -5,6 +5,7 @@ services:
     image: {{ control_plane_image }}
     command: run
     environment:
+      - PGEDGE_IPV4_ADDRESS={{ hostvars[host].peer_ip }}
       - PGEDGE_PEER_ADDRESSES={{ hostvars[host].peer_ip }}
       - PGEDGE_CLIENT_ADDRESSES={{ hostvars[host].external_ip }}
       - PGEDGE_HOST_ID={{ host }}
EOF

# Reset and deploy v0.6.2 to the lima fixture
make reset-lima-fixture FIXTURE_CONTROL_PLANE_IMAGE=ghcr.io/pgedge/control-plane:v0.6.2

# Just to be safe, make sure the clocks are all correct on the fixture
[...]

PR #357, opened 2026-04-20
feat: add state migration to remove ServiceUserRole resources

rshoemaker:
## Summary                                                                                                 
            
  - Add state migration (v1.1.0) to remove `swarm.service_user_role` resources from existing deployments
  - Deletes all `swarm.service_user_role` entries from etcd state regardless of service type                                                                                                                                             
  - Scrubs `swarm.service_user_role` references from all other resources' dependency lists  
  - Bumps state version from 1.0.0 to 1.1.0                                                                                                                                                                                              
                                                                                                             
## Context                                                                                                                                                                                                                             
                                                                                                             
Now that MCP (#338), RAG (#351), and PostgREST (#352) all use `connect_as`, the auto-created `svc_*_ro`/`svc_*_rw` service accounts are no longer needed. This migration cleans up stale state from existing v0.7.0 deployments.       
                                                                                                                                                                                                                                         
Note: orphaned Postgres roles (`svc_*_ro`/`svc_*_rw`) are not dropped from the database — they remain as harmless orphans. Dropping them would require SQL execution during a workflow, which is out of scope for a state migration.
                                                                                                                                                                                                                                         
## Test plan                                                                                               
                                                                                                                                                                                                                                         
  - [x] `TestVersion_1_1_0/removes_service_user_role_resources` — MCP RO+RW removed, deps scrubbed           
  - [x] `TestVersion_1_1_0/removes_service_user_role_resources_across_all_service_types` — MCP, RAG, PostgREST roles removed in single state                                                                                             
  - [x] `TestVersion_1_1_0/no-op_when_no_service_user_role_resources_exist` — safe on clean state                                           
  - [x] `TestVersion_1_1_0/empty_state` — safe on empty state                                                                                                                                                                            
  - [x] Full test suite passes (1061 tests

### DIFF
--- a/server/internal/resource/migrations/1_1_0.go
+++ b/server/internal/resource/migrations/1_1_0.go
@@ -1,0 +1,39 @@
+package migrations
+
+import (
+	"github.com/pgEdge/control-plane/server/internal/ds"
+	"github.com/pgEdge/control-plane/server/internal/resource"
+)
+
+var _ resource.StateMigration = (*Version_1_1_0)(nil)
+
+// Version_1_1_0 removes swarm.service_user_role resources and scrubs
+// references to them from all other resources' dependency lists.
+// Services now use connect_as to reference database_users directly.
+type Version_1_1_0 struct{}
+
+func (v *Version_1_1_0) Version() *ds.Version {
+	return resource.StateVersion_1_1_0
+}
+
+func (v *Version_1_1_0) Run(state *resource.State) error {
+	const serviceUserRoleType resource.Type = "swarm.service_user_role"
+
+	// 1. Delete all service_user_role resources from state
+	delete(state.Resources, serviceUserRoleType)
+
+	// 2. Remove service_user_role from all other resources' dependency lists
+	for _, resources := range state.Resources {
+		for _, data := range resources {
+			filtered := data.Dependencies[:0]
+			for _, dep := range data.Dependencies {
+				if dep.Type != serviceUserRoleType {
+					filtered = append(filtered, dep)
+				}
+			}
+			data.Dependencies = filtered
+		}
+	}
+
+	return nil
+}
--- a/server/internal/resource/migrations/golden_test/TestVersion_1_0_0/empty.json
+++ b/server/internal/resource/migrations/golden_test/TestVersion_1_0_0/empty.json
@@ -1,4 +1,4 @@
 {
-  "version": "1.0.0",
+  "version": "1.1.0",
   "resources": {}
 }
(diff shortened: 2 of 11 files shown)


server/internal/resource/migrations/golden_test/TestVersion_1_0_0/empty.json
```diff
@@ -1,4 +1,4 @@
 {
-  "version": "1.0.0",
+  "version": "1.1.0",
```
jason-lynch: Having these change on every update is going to be a nuisance. We should probably switch `1_0_0_test.go` to not use `NewState`:

```diff
diff --git a/server/internal/resource/migrations/1_0_0_test.go b/server/internal/resource/migrations/1_0_0_test.go
index 703880a2..fb74a858 100644
--- a/server/internal/resource/migrations/1_0_0_test.go
+++ b/server/internal/resource/migrations/1_0_0_test.go
@@ -139,7 +139,10 @@ func TestVersion_1_0_0(t *testing.T) {
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
-			state := resource.NewState()
+			state := &resource.State{
+				Version:   resource.StateVersion_1_0_0,
+				Resources: map[resource.Type]map[string]*resource.ResourceData{},
+			}
 			state.Add(tc.in...)
 
 			migration := &migrations.Version_1_0_0{}
```

jason-lynch (approved):
Nice! This looks good to me. The comment is non-blocking if you're in a hurry. We can fix it on the next update.

server/internal/resource/migrations/golden_test/TestVersion_1_0_0/empty.json
rshoemaker: Fixed.

server/internal/resource/migrations/1_1_0.go
```diff
@@ -0,0 +1,39 @@
...
+	for _, resources := range state.Resources {
+		for _, data := range resources {
+			filtered := data.Dependencies[:0]
+			for _, dep := range data.Dependencies {
+				if dep.Type != serviceUserRoleType {
+					filtered = append(filtered, dep)
+				}
+			}
+			data.Dependencies = filtered
+		}
```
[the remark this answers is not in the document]
rshoemaker: Not valid. ServiceUserRole.TypeDependencies() returns nil, and no other resource lists swarm.service_user_role in its TypeDependencies. TypeDependencies is for declaring that a resource depends on all resources of a given type - nobody depends on all service user roles. The ServiceUserRole references are only in Dependencies (specific resource identifiers), which the migration already handles. No fix needed.

jason-lynch (approved):

PR #362, opened 2026-04-22
fix: move patroni reload operation

jason-lynch:
## Summary

Patroni has an activity loop that they refer to as a 'run cycle'. Each run cycle is `loop_wait` seconds long. When a cycle takes less than `loop_wait` seconds, which is the typical case, Patroni will sleep until that time has elapsed. If the system is degraded and the cycle takes longer than `loop_wait`, it'll sleep for 1ms and log a warning.

When we signal Patroni to reload its configuration, it will perform the reload at the beginning of the next run cycle. This means that under non-degraded conditions, it can take up to `loop_wait` seconds to reload its configuration.

This commit fixes a bug where we were not waiting for the reload to complete. It was working most of the time because we signal Patroni to reload in the `UnitResource`, so the timing worked out that in a lot of cases we would get into `InstanceResource.Create` before the new config took effect, and it would take effect before we evaluated whether a restart was needed.

Rather than doing the wait in the `InstanceResource`, which would happen on every update, I've moved the reload operation to each orchestrator's `PatroniConfig` resource. I've also added a sleep operation here to ensure that `loop_wait` seconds have elapsed before we proceed. This way, we can assume that the new configuration has taken effect in the `InstanceResource` lifecycle methods.

This additional sleep does slow down some operations - especially in Swarm databases because the Patroni Config resource contains the entire instance spec, so there are plenty of conditions where there will be a resource diff even though there isn't a Patroni config change. We'll improve this condition when we adopt the new `common.PatroniConfigGenerator` in the `swarm.patroni_config` resource.

## Testing

This mainly affects the `TestPortChange` E2E on systemd:

```sh
# in terminal one
make dev-lima-run

# in terminal two
use-dev-lima
make test-e2e E2E_RUN=TestPortChange
```

PLAT-548

### DIFF
--- a/lima/roles/stop_dbs/tasks/main.yaml
+++ b/lima/roles/stop_dbs/tasks/main.yaml
@@ -37,3 +37,7 @@
 - name: Reload systemd
   ansible.builtin.command: systemctl daemon-reload
   changed_when: false
+
+- name: Clear journalctl
+  ansible.builtin.command: journalctl --flush --rotate --vacuum-time=1s
+  changed_when: false
--- a/server/internal/utils/utils.go
+++ b/server/internal/utils/utils.go
@@ -57,6 +57,15 @@ func Retry(maxAttempts int, initialDelay time.Duration, f func() error) error {
 	return nil
 }
 
+func SleepContext(ctx context.Context, duration time.Duration) error {
+	select {
+	case <-time.After(duration):
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
 func PointerTo[T any](v T) *T {
 	return &v
 }
--- a/server/internal/database/instance_resource.go
+++ b/server/internal/database/instance_resource.go
@@ -93,6 +93,13 @@ func (r *InstanceResource) Refresh(ctx context.Context, rc *resource.Context) er
 }
 
 func (r *InstanceResource) Create(ctx context.Context, rc *resource.Context) error {
+	if err := r.updateConnectionInfo(ctx, rc); err != nil {
+		return r.recordError(ctx, rc, err)
+	}
+	if err := WaitForPatroniRunning(ctx, r.patroniClient(), 0); err != nil {
+		err = fmt.Errorf("failed to wait for patroni to enter running state: %w", err)
+		return r.recordError(ctx, rc, err)
+	}
 	if err := r.initializeInstance(ctx, rc); err != nil {
 		return r.recordError(ctx, rc, err)
 	}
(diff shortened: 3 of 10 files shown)


lima/roles/stop_dbs/tasks/main.yaml
```diff
@@ -37,3 +37,7 @@
 - name: Reload systemd
   ansible.builtin.command: systemctl daemon-reload
   changed_when: false
+
+- name: Clear journalctl
+  ansible.builtin.command: journalctl --flush --rotate --vacuum-time=1s
+  changed_when: false
```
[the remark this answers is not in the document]
jason-lynch: This Ansible role and the playbooks that use it are for development and testing only. This is intentionally destructive.

server/internal/utils/utils.go
```diff
@@ -57,6 +57,15 @@ func Retry(maxAttempts int, initialDelay time.Duration, f func() error) error {
 	return nil
 }
 
+func SleepContext(ctx context.Context, duration time.Duration) error {
+	select {
+	case <-time.After(duration):
```
tsivaprasad, line 62: Looks it will sleep even if there is a cancellation.
Do you think we can change like below?

```
func SleepContext(ctx context.Context, duration time.Duration) error {
	t := time.NewTimer(duration)
	defer t.Stop()
	select {
	case <-t.C:
		return nil
	case <-ctx.Done():
		return ctx.Err()
	}
}
```

tsivaprasad (approved):
Except for a minor change, everything looks good.

Verification:
```
2026/04/24 00:21:07 initializing cluster
2026/04/24 00:21:21 cluster initialized
=== RUN   TestPortChange
=== PAUSE TestPortChange
=== CONT  TestPortChange
    port_change_test.go:23: [TestPortChange] creating database
    port_change_test.go:56: [TestPortChange] updating database to change ports
    port_change_test.go:85: [TestPortChange] validating that the database is usable
    port_change_test.go:108: [TestPortChange] waiting for replication
    database_test.go:357: [TestPortChange] waiting for replication to catch up on all nodes
    port_change_test.go:112: [TestPortChange] validating that replication is functioning
    fixture_test.go:200: cleaning up database 68cdaaeb-b419-4a30-8784-c58062e4bc35
--- PASS: TestPortChange (110.39s)
PASS
ok  	github.com/pgEdge/control-plane/e2e	124.320s

DONE 1 tests in 124.321s

```

```
2026/04/24 00:29:01 initializing cluster
2026/04/24 00:29:01 cluster initialized
=== RUN   TestUpdateAddNode12
=== PAUSE TestUpdateAddNode12
=== CONT  TestUpdateAddNode12
    db_update_add_node_test.go:34: Creating 1 node database with 1 primary per node and no replicas
    db_update_add_node_test.go:42: Verifying state of node n1
    db_update_add_node_test.go:43: Verifying state of primary
    db_update_add_node_test.go:43: Inserting test data on node n1 primary
    db_update_add_node_test.go:51: Adding new node with zero down time enabled
    db_update_add_node_test.go:53: nodes updated to total number 2
    db_update_add_node_test.go:56: Inserting test data on node n2 primary
    database_test.go:357: [TestUpdateAddNode12] waiting for replication to catch up on all nodes
    db_update_add_node_test.go:269: Verifying data created on peer node n2 is present on n1
    db_update_add_node_test.go:269: Verifying data created on peer node n1 is present on n2
    fixture_test.go:200: cleaning up database 849bf860-43da-42a3-9917-71ef1f7e81ee
--- PASS: TestUpdateAddNode12 (91.81s)
PASS
ok  	github.com/pgEdge/control-plane/e2e	92.335s

DONE 1 tests in 92.336s
```

server/internal/utils/utils.go
jason-lynch: I believe the issue you're referring to was fixed in Go 1.23. The current docs for `time.After` say:

> Before Go 1.23, this documentation warned that the underlying [Timer](https://pkg.go.dev/time@master#Timer) would not be recovered by the garbage collector until the timer fired, and that if efficiency was a concern, code should use NewTimer instead and call [Timer.Stop](https://pkg.go.dev/time@master#Timer.Stop) if the timer is no longer needed. As of Go 1.23, the garbage collector can recover unreferenced, unstopped timers. There is no reason to prefer NewTimer when After will do.

https://pkg.go.dev/time@master#After

PR #358, opened 2026-04-21
feat: remove ServiceUserRole

rshoemaker:
## Summary                                                                                                          

  - Deletes the `ServiceUserRole` resource type and all associated dead code now that all service types use the `connect_as` credential model introduced in PLAT-539/552/553                                                                                                                                                                                      
  - Removes `ServiceUser` type and `GenerateServiceUsername` from the database package                                                                                                                                                                                                                 
  - Removes the `Credentials` field from all structs that carried it (`ServiceInstance`, `ServiceInstanceSpec`,                                                                                                                                                                                        
    `ServiceInstanceSpecResource`, `ServiceContainerSpecOptions`)                                                     
  - Cleans up a dead code path in `generateMCPInstanceResources` where a `serviceType == "rag"` guard would create                                                                                                                                                                                    
    ServiceUserRole resources that could never be reached (RAG dispatches to `generateRAGInstanceResources` before that block)                                                                                                                                                                      
                                                                                                                                                                                                                                         
## Test plan                                                                                                        

  - [X] Unit tests pass: `make test`
  - [X] E2E tests pass
  - [X] Confirm no remaining references to `ServiceUserRole`, `ServiceUser`, or `GenerateServiceUsername` outside the state migration added in #357

PLAT-558

PR #365, opened 2026-04-22
feat: remove pgedge_application built-in roles

rshoemaker:
## Summary                                                                                                                                                                                                                             
                                                                                                                          
  - Removes `pgedge_application` and `pgedge_application_read_only` uilt-in roles from the control plane                                                                                                                                                                                               
  - Deletes `CreateApplicationRole`, `CreateApplicationReadOnlyRole`, and `schemaReadOnly` from `postgres/roles.go`                                                                         
  - Simplifies `CreateBuiltInRoles` to only create `pgedge_superuser`                                                                                                                                                                    
  - Removes application role grants from `GrantBuiltinRolePrivileges`
  - Removes both roles from the `builtinRoles` conflict-check list                                                                                                                                                                       
  - Deletes `docs/development/service-credentials.md` (fully obsolete)                                                    
  - Cleans up stale role references in dev docs                                                                                                                                                                                          
   
## Context                                                                                                                                                                                                                             
                                                                                                            
These roles were created alongside every database but had no remaining consumers:
                                                                                                                                                                                                                                         
  - `pgedge_application_read_only` was used as the PostgREST default anon role — removed in #361 (`db_anon_role` is now required)                                                                                                                                                                         
  - `pgedge_application` lost its last consumer when the `svc_*` service account model was replaced by `connect_as` in #358/#360                                                                                                                                                                      
                                                                                                                                                                                                                                         
This is the cleanup unblocked by those two PRs.                                                                                                                                                                                        
                                                                                                                                                                                                                                         
## Behavior on existing databases                                                                                                                                                                                                      
                                                                                                            
The roles are not dropped from existing databases. They will persist and continue to be replayed to new nodes via `pg_dumpall --roles-only` from the primary. No privileges are revoked. New databases created after this change will never have these roles.                                                                                                                                                                                         
                                                                                                                                                                                                                                         
## Note                                                                                                                                                                                                                         
                                                                                                            
`pgedge_application` and `pgedge_application_read_only` are no longer reserved names. Users may now name their `database_users` entries either of those strings without a validation error.


jason-lynch (approved):
Thanks! We should talk about removing `pgedge_superuser` as well.

PR #366, opened 2026-04-24
fix: embedded etcd should only connect to self

jason-lynch:
## Summary

Fixes a bug where the etcd clients in hosts with embedded etcd were configured to connect to all cluster members that existed when the client was initialized. The correct behavior is that hosts with embedded etcd should only connect to themselves. This was the original intent and functionality, but I changed it while implementing support for remote Etcd. I did a few different implementations of the remote Etcd feature before settling on the current one, and I think this was just an accidental inclusion from a different implementation of that feature.

PLAT-581

## Testing

This PR includes a cluster test for the issue:

```sh
# Using TEST_RERUN_FAILS in case the OS steals our ephemeral ports
make test-cluster TEST_RERUN_FAILS=2 CLUSTER_TEST_RUN=TestRollingAddRemove
```

To run the scenario by hand:

```sh
# start the servers in an uninitialized state
make dev-reset
make dev-watch

# join 4 hosts to the cluster
cp1-req init-cluster
cp2-req join-cluster $(cp1-req get-join-token)
cp3-req join-cluster $(cp1-req get-join-token)

# stop host-1
cp-docker-compose stop host-1

# remove host-1 - remember that this is asynchronous, so watch the logs for it to finish
cp2-req remove-host host-1

# add another host to the cluster
cp4-req join-cluster $(cp2-req get-join-token)
```

Note that our dev setup is slightly different than the scenario in the ticket and our cluster test because hosts 4-6 are in client mode. It still reproduces and validates the issue because the failure occurs in host-2.

### DIFF
--- a/clustertest/add_remove_host_test.go
+++ b/clustertest/add_remove_host_test.go
@@ -3,6 +3,7 @@
 package clustertest
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
@@ -197,4 +198,26 @@ func TestForcedHostRemovalWithDatabase(t *testing.T) {
 	require.NoError(t, err, "database deletion task failed")
 
 	tLog(t, "test completed successfully")
+}
+
+func TestRollingAddRemove(t *testing.T) {
+	// Regression test for PLAT-581
+	t.Parallel()
+
+	// Initialize a four-host cluster, then remove host-1, and add another host.
+	// The order of adds is important because this bug originates in the
+	// endpoint list for host-2+, so we add each host individually.
+	cluster := NewCluster(t, ClusterConfig{
+		Hosts: []HostConfig{
+			{ID: "host-1"},
+		},
+	})
+	cluster.Init(t)
+	for i := 2; i <= 4; i++ {
+		cluster.Add(t, HostConfig{ID: fmt.Sprintf("host-%d", i)})
+		cluster.Init(t)
+	}
+	cluster.Remove(t, "host-1")
+	cluster.Add(t, HostConfig{ID: "host-5"})
+	cluster.Init(t)
 }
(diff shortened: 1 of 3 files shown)


clustertest/add_remove_host_test.go
```diff
@@ -198,3 +199,25 @@ func TestForcedHostRemovalWithDatabase(t *testing.T) {
...
+		},
+	})
+	cluster.Init(t)
+	for i := 2; i <= 4; i++ {
+		cluster.Add(t, HostConfig{ID: fmt.Sprintf("host-%d", i)})
+		cluster.Init(t)
+	}
+	cluster.Remove(t, "host-1")
+	cluster.Add(t, HostConfig{ID: "host-5"})
+	cluster.Init(t)
```
tsivaprasad, line 222: I think we can add a health assertion right after init, and remove the extra init at the end of this test.

tsivaprasad (approved):

clustertest/add_remove_host_test.go
jason-lynch: The `Add` helper here only creates the host and adds it to the client configuration. It does not join it to the cluster. That's why this `Init` call is necessary. `Init` uses the idempotent `InitCluster` method from our client library, which initializes the cluster and joins all members to it. You can take a look at the implementation here: https://github.com/pgEdge/control-plane/blob/main/client/multi.go#L50.

This `Init` call also includes a health assertion because, in the context of tests, we almost always want to verify that the cluster is healthy after modifying its membership. You can see that implementation here: https://github.com/pgEdge/control-plane/blob/main/clustertest/cluster_test.go#L58-L69

I can see why this is confusing, though, because `Add` is asymmetrical with `Remove` since it doesn't modify the cluster membership. I'll open a separate PR to refactor these tests so that `Add` also calls `Init`.

jason-lynch: @tsivaprasad I've opened https://github.com/pgEdge/control-plane/pull/367 to refactor `Add` so that it also calls `Init` to make this less confusing.

PR #369, opened 2026-04-27
chore: resolve ansible facts deprecation notice

jason-lynch:
## Summary

Ansible is deprecating the default setting that populates some ansible facts into top-level `ansible_` variables, such as `ansible_architecture`. This commit replaces our use of these variables with the `ansible_facts` equivalent.

## Testing

This no longer produces a deprecation notice on the latest versions of ansible:

```sh
make dev-lima-deploy
```


tsivaprasad (approved):
Looks Good...!

PR #370, opened 2026-04-29
chore: add missing changelog entries

tsivaprasad:
Fixup missing changelog entries for the next release:

✗ changie batch minor --dry-run
### Added

- Renamed the server binary from `control-plane` to `pgedge-control-plane` to reduce conflicts with other system packages.
- Added PostgREST as a supported service type — Deploy the PostgREST REST API server alongside your database with automatic credential provisioning, upfront schema and role validation, and configurable connection pool settings.
- Released RPM packages for Red Hat-based Linux systems — The Control Plane is now available as native RPM packages for amd64 and arm64, with a bundled systemd service file and SELinux compatibility.
- Added CPU and memory resource limits for systemd-managed database instances — Operators can cap resource consumption per instance using `cpu_quota` and `memory_max` in the instance specification, enforced through systemd cgroup integration.
- Added `connect_as` field for service credentials — Services can now explicitly specify which database user they authenticate as by referencing a `database_users` entry, replacing auto-generated service accounts with direct, auditable credential assignment.
- Added automatic role transfer when expanding a database cluster — PostgreSQL roles created outside the standard `database_users` configuration are now automatically transferred to new nodes when they join a database.
- Extended stable random port assignments to service instances — Ports assigned to MCP, PostgREST, and RAG services are now persisted and reused across restarts and database updates, consistent with the behaviour already in place for database instances.

### Changed

- **Breaking:** The `connect_as` field is now required when creating or updating services of any type (MCP, PostgREST, RAG) — requests that omit this field will be rejected with a validation error.
- **Breaking:** PostgREST services now require `db_anon_role` to be explicitly set in the service configuration — the previously implicit `pgedge_application_read_only` default has been removed.
- Removed the `pgedge_application` and `pgedge_application_read_only` built-in database roles — These roles are no longer created for new databases. The names are no longer reserved and may be used freely for custom database users.

### Fixed

- Fixed port conflicts between services on the same host producing opaque deployment errors — Port conflicts are now detected at creation time and rejected with a clear validation message.
- Fixed `extra_networks` specified in `orchestrator_opts` not being attached to service containers (MCP, PostgREST, RAG).
- Fixed in-place restore operations not completing post-restore steps — The restore configuration is now correctly propagated through the workflow, ensuring database recovery completes fully.
- Fixed the inability to change ports on systemd-managed database instances — Port configuration updates are now applied correctly without requiring a full redeploy.
- Fixed upgrade path from v0.6.2 — Databases created before v0.7.0 that were missing replication slot resources are now automatically repaired during state migration.
- Fixed supporting services not being invalidated when a cluster transitions to systemd orchestration — Service state is now correctly reset when the orchestrator type changes.
- Fixed cryptic Docker errors when database or service identifiers exceeded Swarm name length limits — Identifiers are now validated at the API layer with clear error messages. Database and service IDs are each limited to 36 characters, and their combined length may not exceed 53 characters.
- Fixed embedded etcd clients connecting to all cluster members instead of only their own endpoint — This could cause connectivity issues when cluster membership changed.
- Fixed new replicas failing to bootstrap when added to a node — Configuration updates are now applied to existing instances before new replicas start, ensuring `pg_hba.conf` is correct when replicas begin syncing.

[PLAT-570](https://pgedge.atlassian.net/browse/PLAT-570)

[PLAT-570]: https://pgedge.atlassian.net/browse/PLAT-570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/unreleased/Added-20260331-120000.yaml
+++ b/changes/unreleased/Added-20260331-120000.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: Renamed the server binary from `control-plane` to `pgedge-control-plane` to reduce conflicts with other system packages.
+time: 2026-03-31T12:00:00.000000-04:00
--- a/changes/unreleased/Added-20260403-130000.yaml
+++ b/changes/unreleased/Added-20260403-130000.yaml
@@ -1,3 +1,3 @@
 kind: Added
-body: Released RPM packages for Red Hat-based Linux systems — The Control Plane is now available as native RPM packages for amd64 and arm64, with a bundled systemd service file and SELinux compatibility.
+body: Added preliminary support for systemd as an alternative to Docker Swarm. This feature is currently in "preview" status. You can read more about it in the [systemd page](https://docs.pgedge.com/control-plane/v0-8/installation/systemd) of our docs.
 time: 2026-04-03T13:00:00.000000-04:00
--- a/changes/unreleased/Added-20260403-130000.yaml
+++ b/changes/unreleased/Added-20260403-130000.yaml
@@ -1,3 +1,3 @@
 kind: Added
-body: Added preliminary support for systemd as an alternative to Docker Swarm. This feature is currently in "preview" status. You can read more about it in the [systemd page](https://docs.pgedge.com/control-plane/v0-8/installation/systemd) of our docs.
+body: Added preliminary support for systemd.
 time: 2026-04-03T13:00:00.000000-04:00
--- a/changes/unreleased/Added-20260403-130000.yaml
+++ b/changes/unreleased/Added-20260403-130000.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: Released RPM packages for Red Hat-based Linux systems — The Control Plane is now available as native RPM packages for amd64 and arm64, with a bundled systemd service file and SELinux compatibility.
+time: 2026-04-03T13:00:00.000000-04:00
--- a/changes/unreleased/Added-20260403-130000.yaml
+++ b/changes/unreleased/Added-20260403-130000.yaml
@@ -1,3 +1,3 @@
 kind: Added
-body: Added preliminary support for systemd.
+body: Added preliminary support for systemd as an alternative to Docker Swarm. This feature is currently in "preview" status. You can read more about it in the [systemd page](https://docs.pgedge.com/control-plane/v0-8/installation/systemd) of our docs.
 time: 2026-04-03T13:00:00.000000-04:00
--- a/changes/unreleased/Added-20260406-120000.yaml
+++ b/changes/unreleased/Added-20260406-120000.yaml
@@ -1,3 +1,0 @@
-kind: Added
-body: Added CPU and memory resource limits for systemd-managed database instances — Operators can cap resource consumption per instance using `cpu_quota` and `memory_max` in the instance specification, enforced through systemd cgroup integration.
-time: 2026-04-06T12:00:00.000000-04:00
--- a/changes/unreleased/Added-20260406-120000.yaml
+++ b/changes/unreleased/Added-20260406-120000.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: Added CPU and memory resource limits for systemd-managed database instances — Operators can cap resource consumption per instance using `cpu_quota` and `memory_max` in the instance specification, enforced through systemd cgroup integration.
+time: 2026-04-06T12:00:00.000000-04:00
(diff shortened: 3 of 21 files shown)


changes/unreleased/Added-20260406-120000.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Added
+body: Added CPU and memory resource limits for systemd-managed database instances — Operators can cap resource consumption per instance using `cpu_quota` and `memory_max` in the instance specification, enforced through systemd cgroup integration.
```
jason-lynch: We don't need this changelog entry. systemd support didn't exist in v0.7.0.

changes/unreleased/Fixed-20260416-120000.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Fixed
+body: Fixed in-place restore operations not completing post-restore steps — The restore configuration is now correctly propagated through the workflow, ensuring database recovery completes fully.
```
jason-lynch: This was specific to systemd, it doesn't need to be included in the release notes.

changes/unreleased/Fixed-20260420-120000.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Fixed
+body: Fixed the inability to change ports on systemd-managed database instances — Port configuration updates are now applied correctly without requiring a full redeploy.
```
jason-lynch: This note can be removed.

changes/unreleased/Fixed-20260422-120000.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Fixed
+body: Fixed supporting services not being invalidated when a cluster transitions to systemd orchestration — Service state is now correctly reset when the orchestrator type changes.
```
jason-lynch: What change is this related to? Changing orchestrators would break a lot of things. I don't think we can ever support that operation on an already-initialized host.

changes/unreleased/Fixed-20260424-120000.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Fixed
+body: Fixed cryptic Docker errors when database or service identifiers exceeded Swarm name length limits — Identifiers are now validated at the API layer with clear error messages. Database and service IDs are each limited to 36 characters, and their combined length may not exceed 53 characters.
```
jason-lynch: A few changes for this note:
- This should be labeled as a breaking change
- It also affects host IDs and cluster IDs
- I don't think any internal or external users ever encountered an error from this, so I'm not sure what the cryptic error message is. This was a precautionary change to comply with RFC 1035.

changes/unreleased/Fixed-20260428-120000.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Fixed
+body: Fixed new replicas failing to bootstrap when added to a node — Configuration updates are now applied to existing instances before new replicas start, ensuring `pg_hba.conf` is correct when replicas begin syncing.
```
jason-lynch: This only affected systemd, it does not need a changelog entry.

changes/unreleased/Added-20260403-130000.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Added
+body: Released RPM packages for Red Hat-based Linux systems — The Control Plane is now available as native RPM packages for amd64 and arm64, with a bundled systemd service file and SELinux compatibility.
```
jason-lynch: This should just be a general "added preliminary support for system" note:

```suggestion
body: Added preliminary support for systemd as an alternative to Docker Swarm. This feature is currently in "preview" status. You can read more about it in the [systemd page](https://docs.pgedge.com/control-plane/v0-8/installation/systemd) of our docs.
```

Note that the doc link won't work until we publish our docs for this release.

jason-lynch (commented):
Thanks for doing this! I left a few comments around the systemd portions because that's what I'm familiar with. We don't need granular release notes for systemd-related features and fixes because it didn't exist in the last release.

Could you please also double check that the service-related entries only include changes to things that were present in v0.7.0?

changes/unreleased/Added-20260331-120000.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Added
+body: Renamed the server binary from `control-plane` to `pgedge-control-plane` to reduce conflicts with other system packages.
```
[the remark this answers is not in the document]
jason-lynch: @tsivaprasad I've opened a PR to fix this: https://github.com/pgEdge/control-plane/pull/371

changes/unreleased/Added-20260403-130000.yaml
jason-lynch: @tsivaprasad I don't think this sentence is descriptive enough. Did you see the suggested change in my previous comment?

tsivaprasad: Updated as per suggestion Thanks @jason-lynch

jason-lynch (approved):
Awesome! Thank you for doing this.

PR #375, opened 2026-05-01
docs: promote Supporting Services to GA and fix mkdocs warnings

mmols:
## Summary

- Remove "(Beta)" label from Supporting Services nav and heading
- Add pgEdge RAG Server, PostgREST, cluster tests, and supporting
  service dev guide to mkdocs nav
- Remove manual TOC from running-locally.md that had broken anchors
- Add note that Supporting Services are not yet available on systemd
- Rename "supported service" to "supporting service" in dev guide
- Add changelog entry for Supporting Services promotion

### DIFF
--- a/docs/development/running-locally.md
+++ b/docs/development/running-locally.md
@@ -1,32 +1,5 @@
 # Running the Control Plane locally
 
-- [Running the Control Plane locally](#running-the-control-plane-locally)
-  - [Common prerequisites](#common-prerequisites)
-  - [Developing the Swarm orchestrator](#developing-the-swarm-orchestrator)
-    - [Prerequisites](#prerequisites)
-      - [Configuration](#configuration)
-    - [Running the Control Plane](#running-the-control-plane)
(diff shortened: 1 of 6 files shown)


docs/development/running-locally.md
```diff
@@ -1,32 +1,5 @@
 # Running the Control Plane locally
 
-- [Running the Control Plane locally](#running-the-control-plane-locally)
```
jason-lynch, line 3: Do we have to remove this? I think most people will consume this document in their editor or in GitHub rather than our docs site, and I find the TOC really helpful for navigating it.

jason-lynch: I'd also be open to removing all of our development docs from the docs site.

mmols: Both Github and VS Code have outline views for viewing the table of contents, so it seemed like a consistent experience across all three places to just remove this. Obviously not every editor will have an outline view / markdown navigation features.

I think having the development docs on the docs site is valuable - it promotes the idea of outside contribution.

jason-lynch (approved):

PR #377, opened 2026-05-04
feat: update spec when db versions change

jason-lynch:
## Summary

A database spec can fall out of sync with its running instances when a user upgrades packages outside of our system. We expect this to happen in systemd clusters from typical system package upgrade workflows. So, we need a way to update the database spec when we observe version changes in the running instances.

This PR adds a new monitor that loops over all databases and reconciles the observed versions with those in the instance record and the spec. This monitor uses the existing election system to ensure that only one host runs the reconciliation process.

## Changes

This PR is split into granular commits. I highly recommend reviewing the commits separately rather than looking at the entire diff at once. The most notable of the commits are:

- `fix: use most recently fetched primary instance id`
  - This is an existing bug that I ran into while testing the scenarios in this PR description and the cluster test. Before, we were just picking the first non-empty primary instance ID in the node resource. This could cause a problem if there was a switchover after that primary instance field was updated. This change accounts for that in a backward-compatible way.
- `feat: update host in host monitor`
  - This small change makes it so we update the list of supported versions when the packages change on the host.
- `feat: record versions for replica instances`
  - This was a necessary change for us to determine whether all instances in a node have been updated. It's also a usability improvement to show when a replica has been updated.
- `feat: add service method to reconcile db versions`
  - This commit contains the core logic of this feature.
- `feat: add databases monitor`
  - This is the mechanism that invokes the core logic of this feature.
- `docs: minor postgres version upgrades for systemd`
  - This is a small doc update to describe how we can accommodate minor version upgrades ahead of supporting package management in the control plane API. Externally-invoked major version upgrades are not supported.

## Testing

I've added a new cluster test that exercises this feature. You can also test this by hand using the instructions in this section.

### Using the dev-lima environment

```sh
# reset your environment
make dev-lima-reset
# deploy the fixture updates to add the 'pgedge old' repos
make dev-lima-deploy

# switch to the dev-lima environment
use-dev-lima

# downgrade to the previous version of postgres 18 on hosts 1-3
cp1-ssh 'sudo dnf downgrade -y pgedge-postgresql18'
cp2-ssh 'sudo dnf downgrade -y pgedge-postgresql18'
cp3-ssh 'sudo dnf downgrade -y pgedge-postgresql18'

# start the dev-lima environment in one terminal
make dev-lima-run

# then in another terminal
# switch to the dev-lima environment
use-dev-lima

# initialize the cluster
cp-init

# check the supported versions for hosts 1-3
# you should see 18.2 listed for all three.
cp1-req list-hosts | jq '.hosts[] 
        | select(.id | IN("host-1", "host-2", "host-3"))
        | {id, supported_pgedge_versions}'

# create a database using hosts 1-3
{
  "id": "storefront",
  "spec": {
    "database_users": [
      {
        "username": "admin",
        "password": "password",
        "db_owner": true,
      }
    ],
    "port": 0,
    "patroni_port": 0,
    "nodes": [
      { "name": "n1", "host_ids": ["host-1", "host-2"] },
      { "name": "n2", "host_ids": ["host-3"] }
    ]
  }
}
EOF

# check the instance and spec versions
# you should see that each instance is running 18.2 and the version is set at
# the top-level field in the database spec.
cp1-req get-database storefront | jq '{
    instances : [.instances[] | {id, host_id, node_name, postgres }],
    spec: (.spec | { postgres_version, spock_version, nodes }) }'

# upgrade postgres on host-3
# you might see messages like 'Shared connection to 127.0.0.1 closed.', but they
# should still exit 0 and the message can be ignored.
cp3-ssh 'sudo dnf upgrade -y pgedge-postgresql18'
cp3-ssh 'sudo systemctl try-restart patroni-*'

# within 15 seconds, you should see version 18.3 in the supported versions for
# host-3:
cp1-req list-hosts | jq '.hosts[] 
        | select(.id == "host-3")
        | {id, supported_pgedge_versions}'

# within 30 seconds, you should see a message in the control plane server logs
# that says it detected an updated instance version and an updated node version.
# you should also see the instance is reporting an updated version in the
# databases API and that the database spec has been updated with a version
# override to 18.3 for n2.
cp1-req get-database storefront | jq '{
    instances : [.instances[] | {id, host_id, node_name, postgres }],
    spec: (.spec | { postgres_version, spock_version, nodes }) }'

# next, upgrade host-2
cp2-ssh 'sudo dnf upgrade -y pgedge-postgresql18'
cp2-ssh 'sudo systemctl try-restart patroni-*'

# within 15 seconds, you should see version 18.3 in the supported versions for
# host-2:
cp1-req list-hosts | jq '.hosts[] 
        | select(.id == "host-2")
        | {id, supported_pgedge_versions}'

# within 30 seconds, you should see that the host-2 instance is reporting an
# updated version, however the spec should look the same because n1 is not yet
# fully upgraded.
cp1-req get-database storefront | jq '{
    instances : [.instances[] | {id, host_id, node_name, postgres }],
    spec: (.spec | { postgres_version, spock_version, nodes }) }'

# finally, upgrade host-1
cp1-ssh 'sudo dnf upgrade -y pgedge-postgresql18'
cp1-ssh 'sudo systemctl try-restart patroni-*'

# within 15 seconds, you should see version 18.3 in the supported versions for
# host-1:
cp1-req list-hosts | jq '.hosts[] 
        | select(.id == "host-1")
        | {id, supported_pgedge_versions}'

# within 30 seconds, you should see that the host-1 instance is reporting an
# updated version. You should also see that the spec is updated with version
# 18.3 in the top-level version field, and no more per-node overrides.
cp1-req get-database storefront | jq '{
    instances : [.instances[] | {id, host_id, node_name, postgres }],
    spec: (.spec | { postgres_version, spock_version, nodes }) }'

# perform a no-op update
# we still expect to see some resource updates in the server logs because the
# version number is written into a few resource types. however, these updates
# won't actually change anything.
cp1-req update-database storefront <<EOF | cp-follow-task
{
  "id": "storefront",
  "spec": {
    "database_users": [
      {
        "username": "admin",
        "db_owner": true,
      }
    ],
    "port": 0,
    "patroni_port": 0,
    "nodes": [
      { "name": "n1", "host_ids": ["host-1", "host-2"] },
      { "name": "n2", "host_ids": ["host-3"] }
    ]
  }
}
EOF

# wait 5 seconds to ensure the instance monitor interval has elapsed, then
# re-fetch the updated database. it look identical to the previous get-database
# request with all instances running 18.3 and 18.3 set in the top-level spec
# field.
cp1-req get-database storefront | jq '{
    instances : [.instances[] | {id, host_id, node_name, postgres }],
    spec: (.spec | { postgres_version, spock_version, nodes }) }'

# clean up by running
make dev-lima-reset
```

### Using the compose environment

```sh
# reset your environment
make dev-reset

# ensure you're using the compose environment
use-compose

# start the compose environment in one terminal
make dev-watch

# then in another terminal
# ensure you're using the compose environment
use-compose

# initialize the cluster
cp-init

# create a database using hosts 1-3
{
  "id": "storefront",
  "spec": {
    "database_users": [
      {
        "username": "admin",
        "password": "password",
        "db_owner": true,
      }
    ],
    "port": 0,
    "postgres_version": "18.2",
    "nodes": [
      { "name": "n1", "host_ids": ["host-1", "host-2"] },
      { "name": "n2", "host_ids": ["host-3"] }
    ]
  }
}
EOF

# check the instance and spec versions
# you should see that each instance is running 18.2 and the version is set at
# the top-level field in the database spec.
cp1-req get-database storefront | jq '{
    instances : [.instances[] | {id, host_id, node_name, postgres }],
    spec: (.spec | { postgres_version, spock_version, nodes }) }'

# upgrade postgres on the host-3 instance
docker service update \
    --image=ghcr.io/pgedge/pgedge-postgres:18.3-spock5.0.6-standard-1 \
    postgres-storefront-n2-ant97dj4

# within 30 seconds, you should see a message in the control plane server logs
# that says it detected an updated instance version and an updated node version.
# you should also see the instance is reporting an updated version in the
# databases API and that the database spec has been updated with a version
# override to 18.3 for n2.
cp1-req get-database storefront | jq '{
    instances : [.instances[] | {id, host_id, node_name, postgres }],
    spec: (.spec | { postgres_version, spock_version, nodes }) }'

# upgrade postgres on the host-2 instance
docker service update \
    --image=ghcr.io/pgedge/pgedge-postgres:18.3-spock5.0.6-standard-1 \
    postgres-storefront-n1-9ptayhma

# within 30 seconds, you should see that the host-2 instance is reporting an
# updated version, however the spec should look the same because n1 is not yet
# fully upgraded.
cp1-req get-database storefront | jq '{
    instances : [.instances[] | {id, host_id, node_name, postgres }],
    spec: (.spec | { postgres_version, spock_version, nodes }) }'

# finally, upgrade host-1
docker service update \
    --image=ghcr.io/pgedge/pgedge-postgres:18.3-spock5.0.6-standard-1 \
    postgres-storefront-n1-689qacsi

# within 30 seconds, you should see that the host-1 instance is reporting an
# updated version. You should also see that the spec is updated with version
# 18.3 in the top-level version field, and no more per-node overrides.
cp1-req get-database storefront | jq '{
    instances : [.instances[] | {id, host_id, node_name, postgres }],
    spec: (.spec | { postgres_version, spock_version, nodes }) }'

# perform a no-op update
# we still expect to see some resource updates in the server logs because the
# version number is written into a few resource types. this will trigger a
# patroni reload because the swarm patroni config resource includes the entire
# instance spec. it will also trigger a switchover because the docker CLI sorts
# volume mounts. however, these updates won't change the database versions
# or other configuration.
cp1-req update-database storefront <<EOF | cp-follow-task
{
  "id": "storefront",
  "spec": {
    "database_users": [
      {
        "username": "admin",
        "db_owner": true,
      }
    ],
    "port": 0,
    "nodes": [
      { "name": "n1", "host_ids": ["host-1", "host-2"] },
      { "name": "n2", "host_ids": ["host-3"] }
    ]
  }
}
EOF

# wait 5 seconds to ensure the instance monitor interval has elapsed, then
# re-fetch the updated database. it look identical to the previous get-database
# request with all instances running 18.3 and 18.3 set in the top-level spec
# field.
cp1-req get-database storefront | jq '{
    instances : [.instances[] | {id, host_id, node_name, postgres }],
    spec: (.spec | { postgres_version, spock_version, nodes }) }'

# clean up by running
make dev-reset
```

PLAT-512


server/internal/monitor/provide.go
```diff
@@ -46,7 +52,13 @@ func provideService(i *do.Injector) {
...
 			return nil, err
 		}
-		return NewService(cfg, logger, dbSvc, certSvc, dbOrch, store, hostSvc), nil
+		electionSvc, err := do.Invoke[*election.Service](i)
+		if err != nil {
+			return nil, err
+		}
+
+		candidate := electionSvc.NewCandidate(electionName, cfg.HostID, electionTTL)
+		return NewService(cfg, logger, dbSvc, certSvc, dbOrch, store, hostSvc, candidate), nil
```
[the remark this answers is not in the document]
jason-lynch: This is incorrect. Everything in our `storage` package, including `WatchOp`, use a deferred execution pattern. They don't make any connection to Etcd in their initializers; instead, they connect in an `Exec`, `Start`, or similar method. Creating the candidate without starting it is harmless.

clustertest/external_upgrade_test.go
```diff
@@ -0,0 +1,290 @@
...
+		if testConfig.skipCleanup {
+			tLogf(t, "skipping cleaning for database '%s'", databaseID)
+			return
+		}
+
+		tLogf(t, "cleaning up database '%s'", databaseID)
+
+		resp, err := cluster.Client().DeleteDatabase(ctx, &controlplane.DeleteDatabasePayload{
+			DatabaseID: databaseID,
+		})
```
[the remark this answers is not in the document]
jason-lynch: This is not a concern. The long-running operation here is polling until the task completes, and that is bounded.

docs/installation/systemd.md
```diff
@@ -236,6 +237,37 @@ instructions.
...
+
+Database upgrades are not yet supported via the Control Plane API, but system
+administrators can perform minor Postgres version upgrades by updating the
+packages on each machine. Follow these steps on each host in the cluster:
+
+1. Upgrade Postgres and/or other components using `dnf upgrade`. For example:
+
+    ```sh
+    sudo dnf upgrade pgedge-postgresql18
+    ```
```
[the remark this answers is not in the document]
jason-lynch: This is not a linter or convention that we follow. Our docs use mkdocs syntax.

docs/installation/systemd.md
```diff
@@ -236,6 +237,37 @@ instructions.
...
+
+    ```sh
+    sudo systemctl try-restart <service name>
+    ```
+
+To minimize the risk of downtime, we recommend upgrading one host at a time,
+starting with hosts running replica instances.
+
+After completing the upgrade on all hosts, it may take up to 30 seconds for the
+new versions to be reflected in the database spec in the Control Plane API.
```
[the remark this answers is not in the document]
jason-lynch: This document is written with minimal configuration in mind, whereas that configuration setting is intended for more niche use cases. Your suggested wording is confusing. It's easier for both me, as an author, and readers to understand the way it is currently written.

server/internal/database/reconcile_versions.go
```diff
@@ -0,0 +1,214 @@
...
+		spockVersion := utils.FromPointer(status.Status.SpockVersion)
+		pgEdgeVersion, err := ds.ParsePgEdgeVersion(postgresVersion, spockVersion)
+		if err != nil {
+			continue
+		}
+		pgEdgeVersion, err = pgEdgeVersion.Normalize()
+		if err != nil {
+			continue
+		}
+		if !instance.PgEdgeVersion.Equals(pgEdgeVersion) {
```
mmols, line 143: Is there any scenario where `instance.PgEdgeVersion` could be nil? It seems like it will be populated in the happy path, but i'm thinking when an instance fails.

server/internal/database/reconcile_versions.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+	}
+	if spockMatches && commonSpockVersion != "" && commonSpockVersion != spec.SpockVersion {
+		spec.SpockVersion = commonSpockVersion
+		updatedSpec = spec
+	}
+	if updatedSpec != nil {
+		updatedSpec.NormalizePostgresVersions()
+		if spockMatches && commonSpockVersion != "" {
+			updatedSpec.SpockVersion = commonSpockVersion
```
mmols: What's the purpose of this second guard at the end? It seems unreachable based on what's above.

server/internal/database/reconcile_versions.go
jason-lynch: At least right now there's not, but I'll add some more conditions in case that changes in the future.

server/internal/database/reconcile_versions.go
jason-lynch: It is reachable, but you're right to highlight it because it's redundant. This was me lazily accepting a CodeRabbit suggestion. I'll rearrange these checks and add another test case to ensure we don't normalize the Postgres versions from a Spock-only update.

server/internal/database/reconcile_versions.go
jason-lynch: Fixed in 2a26c915.

server/internal/database/reconcile_versions.go
jason-lynch: Fixed in 2a26c915.

mmols (approved):

PR #380, opened 2026-05-05
feat: enable failsafe mode for single-host nodes

jason-lynch:
## Summary

Patroni's failsafe mode prevents instances from entering read-only mode when they lose their connection to the DCS, as long as they can contact all other Patroni instances in the cluster. There could be edge cases to this behavior when there's more than one instance in the Patroni cluster; however, it's known to be safe if there's only one host in the cluster.

This commit enables failsafe mode when there's only a single host in the node (meaning the Patroni cluster only has a single instance) and disables it otherwise.

This change is backward-compatible and does not need a migration.

## Testing

```sh
# run the compose dev environment
make dev-watch

# then in another terminal, create a database with two nodes: one with two
# hosts and another with one host
{
  "id": "storefront",
  "spec": {
    "database_users": [
      {
        "username": "admin",
        "password": "password",
        "db_owner": true,
      }
    ],
    "port": 0,
    "nodes": [
      { "name": "n1", "host_ids": ["host-1", "host-3"] },
    ]
  }
}
EOF

# in separate terminals, start connections to n1's primary and n2's instance using
# using the interactive prompts
cp-psql

# then, stop the control plane servers with ctrl+c

# wait a few seconds for Patroni to notice that it's lost its connection to Etcd
# then in your psql windows, query pg_is_in_recovery()
select pg_is_in_recovery();

# what you should see is that n1's primary entered recovery mode because it's
# configured with failsafe_mode: false
# n2's primary should _not_ have entered recovery mode and should still be
# writable because it only has a single host.
```

PLAT-545

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - **Breaking:** Database, host, cluster, and service identifiers are now validated to comply with RFC 1035 name requirements — IDs must be 1–36 characters, contain only lowercase letters, digits, and hyphens, and start and end with a letter or digit. The combined length of a database ID and service ID may not exceed 53 characters.
 - Removed the `pgedge_application` and `pgedge_application_read_only` built-in database roles — These roles are no longer created for new databases. The names are no longer reserved and may be used freely for custom database users.
 - Promote Supporting Services from beta to generally available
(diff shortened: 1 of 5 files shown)


jason-lynch:
@coderabbitai review

mmols (approved):

PR #392, opened 2026-05-26
feat: add Image and ResolvedImage fields to SwarmOpts

tsivaprasad:
## Summary
This PR adds `Image` and `ResolvedImage` fields to `SwarmOpts` and implements precedence-based image resolution in the Swarm orchestrator. This provides the foundation for user-controlled image overrides and stable image pinning.

## Changes
* Added `Image` (user override, never managed by CP) and `ResolvedImage` (CP-managed) fields to `SwarmOpts` in `server/internal/database/spec.go`, serialized under `orchestrator_opts.docker`.
* Updated `SwarmOpts.Clone()` to correctly copy both new fields.
* Extracted `resolveInstanceImages()` in the Swarm orchestrator to implement precedence-based image resolution: `Image` → `ResolvedImage` → manifest lookup (lazy backfill).
* When `Image` is explicitly set, the manifest is bypassed entirely, allowing unknown or custom versions to work through user overrides.

## Testing
Verification:
- Created cluster
- Created DB using [create_db.json](https://github.com/user-attachments/files/28251194/create_db.json)
- Confirm Docker service resolved the correct manifest image
```
docker service ls
ID             NAME                              MODE         REPLICAS   IMAGE                                                       PORTS
l2837hwwvoy8   postgres-storefront-n1-9ptayhma   replicated   1/1        ghcr.io/pgedge/pgedge-postgres:17.9-spock5.0.6-standard-2   
c6wg777c5rig   postgres-storefront-n1-689qacsi   replicated   1/1        ghcr.io/pgedge/pgedge-postgres:17.9-spock5.0.6-standard-2   
seysxqetlp0s   postgres-storefront-n1-ant97dj4   replicated   1/1        ghcr.io/pgedge/pgedge-postgres:17.9-spock5.0.6-standard-2 
```

## Checklist

- [x] Tests added or updated

[PLAT-596](https://pgedge.atlassian.net/browse/PLAT-596)

[PLAT-596]: https://pgedge.atlassian.net/browse/PLAT-596?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/internal/orchestrator/swarm/orchestrator.go
```diff
@@ -171,10 +171,41 @@ func ServiceInstanceName(databaseID, serviceID, hostID string) string {
...
+func (o *Orchestrator) resolveInstanceImages(spec *database.InstanceSpec) (*Images, error) {
+	var swarmOpts *database.SwarmOpts
+	if spec.OrchestratorOpts != nil {
+		swarmOpts = spec.OrchestratorOpts.Swarm
+	}
+
+	switch {
+	case swarmOpts != nil && swarmOpts.Image != "":
+		return &Images{PgEdgeImage: swarmOpts.Image}, nil
+	case swarmOpts != nil && swarmOpts.ResolvedImage != "":
```
jason-lynch, line 197: What happens when `spec.PgEdgeVersion` changes? Do we need to check that the resolved image still satisfies the `PgEdgeVersion` or are you planning to handle that elsewhere?

server/internal/orchestrator/swarm/orchestrator.go
```diff
@@ -171,10 +171,41 @@ func ServiceInstanceName(databaseID, serviceID, hostID string) string {
...
+			spec.OrchestratorOpts.Swarm = &database.SwarmOpts{}
+		}
+		spec.OrchestratorOpts.Swarm.ResolvedImage = manifested.PgEdgeImage
+		return &Images{PgEdgeImage: manifested.PgEdgeImage}, nil
+	}
+}
+
 func (o *Orchestrator) instanceResources(spec *database.InstanceSpec, scripts database.Scripts) (*database.InstanceResource, []resource.Resource, []resource.Resource, error) {
-	images, err := o.versions.GetImages(spec.PgEdgeVersion)
+	images, err := o.resolveInstanceImages(spec)
```
jason-lynch, line 234: `resolveInstanceImages` modifies the spec, but it's not written back to storage here, so `ResolvedImage` isn't persisted in this implementation.

We could write the spec back to storage in this method, however I think it would be cleaner to do this operation in `database.Service.ReconcileInstanceSpec` since it's job is to read, modify, and store the instance spec, and it gets called before `database.Orchestrator.GenerateInstanceResources` which calls this method.

I would consider adding a new method to the `database.Orchestrator` interface, like `ReconcileInstanceSpec(old, new *database.InstanceSpec) error`. The systemd implementation would be a no-op, but for Swarm you could call `o.resolveInstanceImages(new)`. Even if we don't use the `old` param just yet, I think it's a sensible to include it in this type of method.

Also - please make sure to include the new fields in `database.InstanceSpec.CopySettingsFrom`. the `InstanceSpec`s are regenerated from the database spec every time, so anything that can't be generated and that we want to keep from the previous spec needs to be included in `CopySettingsFrom`. Otherwise, it won't be maintained between updates.

tsivaprasad: Addressed except below:
For `CopySettingsFrom`, I initially tried copying `ResolvedImage`, but it caused `TestExternalUpgrade` to fail. The issue is that the monitor updates `PgEdgeVersion` directly in storage without going through `Service.ReconcileInstanceSpec`, which leaves the stored spec with a stale `ResolvedImage` pointing to the previous version's image. When `CopySettingsFrom` carries that value forward, a subsequent no-op update ends up using the old image and unintentionally rolls back the externally upgraded instance.

Since `ReconcileInstanceSpec` always resolves the image fresh from the manifest (an in-memory lookup with no network overhead), `ResolvedImage` is kept up to date automatically and does not need to be copied in `CopySettingsFrom`.

jason-lynch (approved):
Awesome! I was little confused for a second because it looks like the `ManifestLoader` PR got merged into this one, but I can see that it's still unused.

tsivaprasad:
> Awesome! I was little confused for a second because it looks like the `ManifestLoader` PR got merged into this one, but I can see that it's still unused.

Thanks @jason-lynch 
The manifest-related changes were introduced as part of PR https://github.com/pgEdge/control-plane/pull/394. They're approved and merged to this branch as part of other ticket and is foundational groundwork — future PRs will wire up the ManifestLoader into the orchestrator to replace the current hardcoded NewVersions(). This is incremental/stacked development, so these pieces land ahead of the PR that actually activates them.

PR #394, opened 2026-05-29
feat: add ManifestLoader for version manifest

tsivaprasad:
## Summary
This PR implements `ManifestLoader`, the component responsible for loading, caching, and periodically refreshing the version manifest. It ensures the Control Plane can start reliably by falling back through a hierarchy of sources: r`emote URL → disk cache → embedded manifest`, allowing startup even when network or disk resources are unavailable.

## Changes
* Added `ManifestLoader` in `server/internal/orchestrator/swarm/manifest_loader.go` to load and manage the version manifest.

  * Resolves manifests using a fallback chain: `manifest_path` (local file) → remote URL → disk cache → embedded manifest.
  * Logs warnings at each fallback step, including the source and reason for the fallback.
  * Validates manifest structure, rejecting malformed JSON or unsupported `schema_version` values before attempting the next fallback source.
  * Runs a background refresh every hour when using a remote URL; refresh failures preserve the currently loaded in-memory manifest.
  * Produces `*Versions` and `*ServiceVersions` objects compatible with the existing `NewVersions()` and `NewServiceVersions()` constructors.

* Added optional `manifest_url` and `manifest_path` configuration fields to `DockerSwarm` in `config/config.go`.

* Embedded `version-manifest.json` into the binary using `//go:embed`, providing a guaranteed fallback manifest for Control Plane startup.

## Testing

- Integration test against a real local file server
```
cd server/internal/orchestrator/swarm && python3 -m http.server 9090

MANIFEST_TEST_URL=http://localhost:9090/version-manifest.json \
  go test -v -run TestManifestLoader_RealURL \
  ./server/internal/orchestrator/swarm/...
```
```
=== RUN   TestManifestLoader_RealURL
{"level":"info","component":"manifest_loader","source":"url:http://localhost:9090/version-manifest.json","time":"2026-05-29T17:37:41+05:30","message":"version manifest loaded"}
    manifest_loader_test.go:495: loaded 15 supported versions; default=18.4
{"level":"info","component":"manifest_loader","url":"http://localhost:9090/version-manifest.json","time":"2026-05-29T17:37:41+05:30","message":"version manifest refreshed"}
    manifest_loader_test.go:507: refresh OK; default still 18.4
--- PASS: TestManifestLoader_RealURL (0.02s)
PASS
ok      github.com/pgEdge/control-plane/server/internal/orchestrator/swarm      0.287s
```

## Checklist

- [x] Tests added 

## Notes for Reviewers

- DefaultManifestURL is a placeholder — TODO(PLAT-598) marks the line. 
- Cache path uses `cfg.DataDir/manifests/version-manifest-cache.json` (consistent with how all other runtime data is stored).


tsivaprasad:
@coderabbitai review

server/internal/orchestrator/swarm/manifest_loader.go
```diff
@@ -0,0 +1,394 @@
...
+
+//go:embed version-manifest.json
+var embeddedManifest []byte
+
+const (
+	// DefaultManifestURL is the URL used when no manifest_url is configured.
+	// TODO(PLAT-598): Replace with the real  URL once the hosting location is confirmed.
+	DefaultManifestURL = "https://download.pgedge.com/manifests/version-manifest.json"
+
+	manifestCacheFilename  = "version-manifest-cache.json"
```
jason-lynch: These defaults should live in `config.go`. We have default values for each config struct and they get combined with the user-specified configuration. These would go in the `defaultDockerSwarm` variable in `config.go`.

server/internal/orchestrator/swarm/manifest_loader.go
```diff
@@ -0,0 +1,394 @@
...
+}
+
+// NewManifestLoader creates and starts a ManifestLoader.  It loads the
+// manifest synchronously before returning so callers always get a valid
+// *Versions / *ServiceVersions immediately.  Background refresh (if
+// applicable) is started as a goroutine tied to ctx.
+func NewManifestLoader(ctx context.Context, cfg config.Config, logger zerolog.Logger, opts ...manifestLoaderOption) *ManifestLoader {
+	m := &ManifestLoader{
+		cfg:       cfg,
+		logger:    logger.With().Str("component", "manifest_loader").Logger(),
```
jason-lynch: We have a different pattern for setting the `component` field. Each component has a constant in `logging/factory.go`. Then, your initializer would take a `logging.Factory` rather than a `zerolog.Logger`, and then it would do:

```go
        logger: loggerFactory.Logger(logging.ComponentManifestLoader)
```

You can look at some of the other components, like `database.Service` as an example.

For your tests, we have a helper in the `testutils` package: https://github.com/pgEdge/control-plane/blob/4f0cd6a7c48d2d12afe9bced5643de13d11ecf3b/server/internal/testutils/logger.go#L23

You can find examples of how its used in a few tests. The simplest is just passing `testutils.LoggerFactory(t)` to your initializer, like we do here: https://github.com/pgEdge/control-plane/blob/4f0cd6a7c48d2d12afe9bced5643de13d11ecf3b/server/internal/etcd/remote_test.go#L33

server/internal/orchestrator/swarm/manifest_loader.go
```diff
@@ -0,0 +1,394 @@
...
+		}
+	}
+
+	// 4. Embedded binary manifest.
+	m.logger.Warn().Msg("using embedded manifest; consider checking connectivity to manifest URL")
+	return embeddedManifest, "embedded"
+}
+
+// manifestURL returns the configured URL, or DefaultManifestURL if unset.
+func (m *ManifestLoader) manifestURL() string {
```
jason-lynch: This method won't be necessary one you move the default value into `config.go`.

server/internal/orchestrator/swarm/manifest_loader.go
```diff
@@ -0,0 +1,394 @@
...
+
+	if cfg.DockerSwarm.ManifestPath == "" {
+		go m.refreshLoop(ctx)
+	}
+
+	return m
+}
+
+// Versions returns the current parsed *Versions.
+func (m *ManifestLoader) Versions() *Versions {
```
jason-lynch: Non-blocking: I think it would be better not to use a pointer for `Versions`. This is an internal package, so it's not a real worry since we control how it's used, but returning a `*Versions` makes it possible for callers to modify `ManifestLoader`'s copy of the value, and they can do so in a non-thread-safe way. We make a similar choice to use plain structs in a few others places, like in `config.Manager`.

server/internal/orchestrator/swarm/manifest_loader.go
```diff
@@ -0,0 +1,376 @@
...
+	if err != nil {
+		// Embedded manifest is malformed — this is a build-time error.
+		panic(fmt.Sprintf("manifest_loader: failed to parse manifest from %s: %v", src, err))
+	}
+	m.mu.Lock()
+	m.versions = v
+	m.svcVersions = sv
+	m.mu.Unlock()
+	m.logger.Info().Str("source", src).Msg("version manifest loaded")
+}
```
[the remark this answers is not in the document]
jason-lynch: @tsivaprasad Is this a valid comment? I don't think we should panic if the local manifest is malformed, but it probably does make sense to return an error and not use the fallback chain so that the user can see that there's an error. We should also consider how to inform the user that their custom manifest URL is broken.

Maybe, rather than framing this as a single set of fallbacks, it makes more sense to think of it separate chains:

- The default chain:
  - Remote manifest -> cached manifest -> embedded manifest -> failure
- The custom remote chain
  - Remote manifest -> cached manifest -> failure
  - Note that this skips the embedded manifest
  - We can detect a custom remote by comparing the configured manifest URL to the default one
- The custom local chain
  - Local manifest -> failure

We should also think about how to invalidate the cached manifest when the manifest URL changes. Maybe the cache filename should be a hash of the manifest URL?

server/internal/orchestrator/swarm/manifest_loader.go
```diff
@@ -0,0 +1,394 @@
...
+	manifestCacheFilename  = "version-manifest-cache.json"
+	refreshInterval        = time.Hour
+	fetchTimeout           = 10 * time.Second
+	supportedSchemaVersion = 1
+)
+
+// versionManifest is the top-level structure of version-manifest.json.
+type versionManifest struct {
+	SchemaVersion int                        `json:"schema_version"`
+	Images        map[string]json.RawMessage `json:"images"`
```
jason-lynch: What's the rationale for using `map[string]json.RawMessage` instead of a dedicated struct, like:

```go
type Images struct {
	Postgres  []manifestPostgresEntry `json:"postgres"`
	PostgREST []manifestServiceEntry  `json:"postgrest"`
	MCP       []manifestServiceEntry  `json:"mcp"`
	RAG       []manifestServiceEntry  `json:"rag"`
}
```

I think that would simplify some of the parsing logic below, right?

server/internal/orchestrator/swarm/manifest_loader.go
```diff
@@ -0,0 +1,394 @@
...
+}
+
+// writeCache persists data to the cache path, creating parent directories as
+// needed.  Errors are logged but not returned (caching is best-effort).
+func (m *ManifestLoader) writeCache(data []byte) error {
+	if err := os.MkdirAll(filepath.Dir(m.cachePath), 0o755); err != nil {
+		m.logger.Warn().Err(err).Str("path", m.cachePath).Msg("failed to create manifest cache directory")
+		return err
+	}
+	if err := os.WriteFile(m.cachePath, data, 0o644); err != nil {
```
jason-lynch: Could you please change these to `0o700` and `0o600`? That's what we use for the data directory and its other contents.

```suggestion
	if err := os.MkdirAll(filepath.Dir(m.cachePath), 0o700); err != nil {
		m.logger.Warn().Err(err).Str("path", m.cachePath).Msg("failed to create manifest cache directory")
		return err
	}
	if err := os.WriteFile(m.cachePath, data, 0o600); err != nil {
```

tsivaprasad:
@coderabbitai review

server/internal/orchestrator/swarm/manifest_loader.go
```diff
@@ -0,0 +1,426 @@
...
+}
+
+// resolveURL tries to load the manifest from the configured URL, falling back
+// to the disk cache.  For the default URL it also falls back to the embedded
+// manifest.  For a custom URL it returns an error if all sources fail so the
+// operator knows their configuration is broken.
+func (m *ManifestLoader) resolveURL() ([]byte, string, error) {
+	u := m.cfg.DockerSwarm.ManifestURL
+
+	data, err := m.fetchURL(context.Background(), u)
```
jason-lynch: It seems like there's a context available in the methods that lead to this one. Should we propagate that context into this method?

server/internal/orchestrator/swarm/manifest_loader.go
```diff
@@ -0,0 +1,426 @@
...
+// load performs the initial synchronous load.
+func (m *ManifestLoader) load() error {
+	data, src, err := m.resolve()
+	if err != nil {
+		return err
+	}
+	v, sv, err := m.parseManifestData(data)
+	if err != nil {
+		if src == "embedded" {
+			// The embedded manifest is corrupt — this is a build-time error.
```
jason-lynch, line 177: You could also add a test that validates the embedded manifest. That way we would see the issue pretty early in our CI pipeline.

jason-lynch (approved):
This looks good to me! I left a few minor suggestions, but nothing blocking.

PR #396, opened 2026-06-03
chore: add replace directives for mergo in go.mod

jason-lynch:
## Summary

The domain registration for dario.cat has expired, which has broken our licenses check and breaks build anywhere that doesn't have this module cached. There's a potential security risk if the author does not re-register this domain because a malicious party could capture this domain and redirect to a modified copy of the module.

Luckily, this is just a vanity URL, and the module is hosted on GitHub. This commit fixes the issue by adding `replace` directives to our go.mod file. Note that we will need to manually keep the replaced version in sync with our dependencies.


tsivaprasad (approved):
Looks good.

PR #399, opened 2026-06-04
feat: add pg_hba_conf/pg_ident_conf to db spec

moizpgedge:
# PR: feat: add pg_hba_conf/pg_ident_conf to db spec

Add user-managed `pg_hba_conf` and `pg_ident_conf` fields to the database spec, at both the database and per-node level, as arrays of strings (matching Patroni's `pg_hba`/`pg_ident` contract).

- Per-node entries are **prepended** to database-level entries, giving node rules first-match priority while keeping the database-level list as the baseline. Empty/unset is identical to today's behavior.
- Add a pg_hba/pg_ident line parser (`postgres/hba`) and minimal spec-time validation: reject `hba_file`/`ident_file` in `postgresql_conf` (they would make the entries ineffective) and lines that fail to parse, surfacing the offending line. No enum-style restriction on auth methods.

Scope: spec contract, parser, and validation only. The orchestrator generators do not yet emit these entries into the generated `pg_hba.conf`/`pg_ident.conf`; that lands in a follow-up.

See internal-design-docs: pg_hba / pg_ident User Configuration.

[PLAT-615](https://pgedge.atlassian.net/browse/PLAT-615)

---

## Summary

Adds `pg_hba_conf` and `pg_ident_conf` to the database spec (database-level and per-node) so operators can supply their own authentication rules. This PR covers the spec contract, a pg_hba/pg_ident parser, and validation only — the entries are accepted, validated, stored, and round-tripped, but are not yet emitted into the generated config (follow-up ticket).

## Changes

- **API design** (`api/apiv1/design/database.go`): add `pg_hba_conf` and `pg_ident_conf` (`ArrayOf(String)`) to `DatabaseSpec` and `DatabaseNodeSpec`; regenerated Goa code and OpenAPI.
- **Domain model** (`server/internal/database/spec.go`): add both fields to `Spec`, `Node`, and `InstanceSpec` (and their `Clone()` helpers); prepend-merge node-level entries ahead of database-level entries in `NodeInstances()`.
- **API conversion** (`server/internal/api/apiv1/convert.go`): map the new fields in both directions, at the database and node level.
- **Parser** (`server/internal/postgres/hba/parse.go`): new `ParseEntry` (pg_hba) and `ParseIdent` (pg_ident) with a quote-aware tokenizer; structural validation only (connection type + field count), no auth-method enum restriction.
- **Validation** (`server/internal/api/apiv1/validate.go`): reject `hba_file`/`ident_file` GUCs and unparseable lines (surfacing the offending line); wired into both `validateDatabaseSpec` and `validateNode`, so it runs on create and update.
- **Tests**: parser unit tests, prepend-merge test, and validation cases (valid entries, unparseable lines, forbidden GUCs).

## Testing

```sh
# build / static checks
go build ./...
go vet ./server/internal/...
make lint                       # golangci-lint: 0 issues

# unit tests (incl. golden tests — no regression)
go test ./server/internal/postgres/hba/... \
        ./server/internal/database/... \
        ./server/internal/api/apiv1/...

# API codegen is in sync
make -C api generate            # only the new fields change (Go); OpenAPI example churn is goa faker noise
```

Manual verification against a local dev cluster (restish):

- **Create** with `pg_hba_conf`/`pg_ident_conf` (db-level + node-level) → accepted; `get-database` round-trips the exact entries.
- **Update** an existing database's entries → accepted (`modifying` → `available`); changes round-trip.
- **Validation** on both create and update: invalid pg_hba line, invalid pg_ident line, and `hba_file`/`ident_file` GUC each return `400 invalid_input` with the offending line/key; a non-standard auth method (`ldap2`) is accepted.
- Confirmed the running container's `pg_hba.conf` is **unchanged** (entries not yet emitted — expected for this scope).

## Checklist

- [ ] Documentation updated (if needed) — deferred to the follow-up ticket, when the entries take effect
- [x] Issue is linked (branch name or URL in PR description)
- [ ] [Changelog](https://github.com/pgEdge/control-plane/blob/main/docs/development/development.md#pull-requests) entry added for user-facing behavior changes — deferred until the feature is user-effective (entries emitted into the file)
- [ ] Breaking changes (if any) are clearly called out in the PR description — none; new fields are optional and additive

## Notes for Reviewers

- **Scope boundary:** this is ticket 1 of the feature. The fields are validated and stored but the generators (`swarm`/`common`) do not yet write them into `pg_hba.conf`/`pg_ident.conf`, and there is no reload yet. That is the follow-up ticket.
- **Generated diff:** `api/apiv1/gen/http/openapi3.*` shows a large diff because goa's example faker reseeds and reshuffles every placeholder when the design changes. It is deterministic and expected (a clean design regenerates to zero diff). Happy to split the regenerated `gen/**` into its own commit if that helps review.
- **Validation is intentionally minimal** per the design doc (admins own these files): only `hba_file`/`ident_file` and unparseable lines are rejected — no auth-method enum, no `trust`-on-broad-CIDR block, no duplicate or `map=` cross-ref checks. PostgreSQL surfaces deeper errors at reload time.
- **Backward compatible:** empty/unset fields produce byte-identical output to today (the new struct fields use `omitempty`; golden tests are unchanged).

[PLAT-615]: https://pgedge.atlassian.net/browse/PLAT-615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Result:-
```
moiz@PGEDGE-ML-Moiz work %  restish control-plane-local-1 get-database pghba
HTTP/1.1 200 OK
Content-Length: 943
Content-Type: application/json
Date: Thu, 04 Jun 2026 14:16:41 GMT

{
  created_at: "2026-06-04T14:15:37Z"
  id: "pghba"
  instances: [
    {
      created_at: "2026-06-04T14:15:40Z"
      host_id: "host-1"
      id: "pghba-n1-689qacsi"
      node_name: "n1"
      postgres: {
        patroni_state: "running"
        role: "primary"
        version: "18.4"
      }
      spock: {
        read_only: "off"
        version: "5.0.8"
      }
      state: "available"
      status_updated_at: "2026-06-04T14:16:39Z"
      updated_at: "2026-06-04T14:16:04Z"
    }
  ]
  spec: {
    database_name: "testdb"
    database_users: [
      {
        attributes: ["LOGIN", "SUPERUSER"]
        db_owner: true
        username: "admin"
      }
      {
        attributes: ["LOGIN"]
        db_owner: false
        username: "myapp_user"
      }
    ]
    nodes: [
      {
        host_ids: ["host-1"]
        name: "n1"
        pg_hba_conf: ["host testdb myapp_user 10.0.0.0/8 scram-sha-256"]
      }
    ]
    pg_hba_conf: [
      "hostssl all myapp_user 203.0.113.0/24 scram-sha-256"
      "host all all 0.0.0.0/0 ldap2"
    ]
    pg_ident_conf: ["ssl_users CN=alice,O=example alice"]
    postgres_version: "18.4"
    spock_version: "5"
  }
  state: "available"
  updated_at: "2026-06-04T14:15:37Z"
}
moiz@PGEDGE-ML-Moiz work % 

```


server/internal/database/spec.go
```diff
@@ -716,6 +728,8 @@ func (s *Spec) NodeInstances() ([]*NodeInstances, error) {
 				BackupConfig:     overridableValue(s.BackupConfig, node.BackupConfig),
 				RestoreConfig:    effectiveRestore,
 				PostgreSQLConf:   overridableMapValue(s.PostgreSQLConf, node.PostgreSQLConf),
+				PgHbaConf:        prependEntries(s.PgHbaConf, node.PgHbaConf),
```
jason-lynch: Is there a reason to use a custom helper function instead of `slices.Concat`?:

```go
PgHbaConf:        slices.Concat(s.PgHbaConf, node.PgHbaConf),
PgIdentConf:      slices.Concat(s.PgIdentConf, node.PgIdentConf),
```

server/internal/postgres/hba/parse.go
```diff
@@ -0,0 +1,201 @@
...
+//
+// Validation is intentionally minimal. Per the design, the control plane does
+// not second-guess the operator: there are NO enum-style restrictions on the
+// auth method, no blocking of trust on broad CIDRs, no duplicate detection, and
+// no map= cross-reference checks. We only reject lines that are not
+// recognizable pg_hba/pg_ident entries at all (unknown connection type, too few
+// fields, unterminated quotes). PostgreSQL surfaces deeper configuration errors
+// (bad auth methods, addresses, options) at reload time via the task logs.
+
+var entryTypes = map[EntryType]struct{}{
```
jason-lynch: We should use `ds.Set` here rather than a plain `map`. Internally, `ds.Set` is a `map[comparable]bool`, but it has a bunch of methods for common set operations. So here you would do:

```go
var entryTypes = ds.NewSet(
	EntryTypeLocal,
	EntryTypeHost,
	EntryTypeHostSSL,
	EntryTypeHostNoSSL,
	EntryTypeHostGSSEnc,
	EntryTypeHostNoGSSEnc,
	EntryTypeInclude,
	EntryTypeIncludeIfExists,
	EntryTypeIncludeDir,
)
```

And then to check whether something is a member of the set, you would do:

```go
if entryTypes.Has(typ) {
    // ...
}
```

server/internal/postgres/hba/parse.go
```diff
@@ -0,0 +1,201 @@
...
+
+func isIncludeType(t EntryType) bool {
+	return t == EntryTypeInclude || t == EntryTypeIncludeIfExists || t == EntryTypeIncludeDir
+}
+
+// isBareIP reports whether s is a plain IP address with no CIDR suffix. It is
+// used only to disambiguate the optional separate IP-mask form of a host entry
+// ("ADDRESS MASK METHOD"); it is not a validity check on the address.
+func isBareIP(s string) bool {
+	return !strings.Contains(s, "/") && net.ParseIP(s) != nil
```
jason-lynch: Is the `string.Contains` necessary here? In my testing, `net.ParseIP` already returns `nil` for strings with a CIDR suffix, like: https://go.dev/play/p/XBN1M7uw0uo

server/internal/postgres/hba/parse.go
```diff
@@ -0,0 +1,201 @@
...
+// unquoted '#' comment. It returns an error for an unterminated quote.
+func tokenize(line string) ([]string, error) {
+	var (
+		tokens   []string
+		buf      strings.Builder
+		inQuote  bool
+		hasToken bool
+	)
+	flush := func() {
+		if hasToken {
```
jason-lynch: I think you could replace this condition and the `hasToken` variable with:


```suggestion
		if buf.Len() > 0 {
```

server/internal/postgres/hba/parse_test.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			got, err := ParseEntry(tc.line)
+			if err != nil {
+				t.Fatalf("ParseEntry(%q) returned error: %v", tc.line, err)
+			}
+			if got != tc.want {
+				t.Errorf("ParseEntry(%q)\n  got:  %#v\n  want: %#v", tc.line, got, tc.want)
+			}
```
jason-lynch: We use a library called `testify` to make these types of assertions easier to write. It produces good test failure messages, so I hardly ever customize them. It has two main packages `assert` which has functions that will fail the test, but allow it to continue, and `require` which has functions that will fail the test immediately:

```suggestion
			got, err := ParseEntry(tc.line)
			require.NoError(t, err)
			require.Equal(t, tc.want, got)
```

You'll see a lot of different usage examples of both of `assert` and `require` throughout our code.

server/internal/postgres/hba/parse_test.go
```diff
@@ -0,0 +1,151 @@
...
+		{"missing auth method", "host all all 0.0.0.0/0"},
+		{"local missing method", "local all all"},
+		{"include without path", "include"},
+		{"include with too many args", "include a b"},
+		{"empty", ""},
+		{"unterminated quote", `host "all all 0.0.0.0/0 md5`},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			if _, err := ParseEntry(tc.line); err == nil {
```
jason-lynch: Same as above, I recommend using `testify` here. It has a few types of assertions for errors. The ones that I use the most frequently are either a plain assertion that an error exists:

```go
require.Error(t, err)
```

An assertion for a sentinel error, e.g.:

```go
require.ErrorIs(t, err, storage.ErrNotFound)
```

Or an assertion for part of an error message:

```go
// where tc.err is a string with part of the error message, like "unknown connection type"
require.ErrContains(t, err, tc.err)
```

For these particular tests, I'd probably add an `err string` field to the test struct and use `ErrContains`.

server/internal/postgres/hba/parse_test.go
```diff
@@ -0,0 +1,151 @@
...
+		},
+		{
+			name: "include directive",
+			line: "include_if_exists /etc/pg_ident_extra.conf",
+			want: IdentEntry{Include: EntryTypeIncludeIfExists, IncludePath: "/etc/pg_ident_extra.conf"},
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			got, err := ParseIdent(tc.line)
```
jason-lynch, line 123: Same comments about `testify` for this test suite and `TestParseIdentErrors`.

jason-lynch (commented):
Nice work! All of my comments are really minor, and the parser implementation is neat!

Since you're adding the API fields here we'll want to hold off on the release until this feature is complete. I'm on board with that since we have a few other in-flight changes we want to include in the release as well.

server/internal/database/spec.go
moizpgedge: not really. 
I can change it to use slices.concat

server/internal/postgres/hba/parse.go
moizpgedge: You right for this net.ParseIo return nils

jason-lynch (approved):
Awesome! Thank you for making those changes.

PR #407, opened 2026-06-16
feat: consolidate duplicate port validation

jason-lynch:
## Summary

We were validating duplicate ports in a few spots, each with different error message shapes, and neither could catch duplicate ports from nodes deployed to the same host.

This PR consolidates and improves those checks so that we'll catch duplicate ports from Postgres, Patroni, or other services on the same host.

## Changes

This PR is split into two commits:

- `refactor: add validation package` - refactors the validation error and path utilities to a new `validation` package. This starts to clean up our validation code and lets me use a `Unique` validator I originally implemented on another branch.
- `feat: consolidate duplicate port validation` - Adds a `Unique` validator to the `validation` package and uses it to validate unique ports across hosts. This replaces the disparate duplicate-port validation we had for both nodes and services.

## Testing

```sh
# This request would be accepted on main, but would result in an error
# You can use any environment (compose, dev-lima, lima, etc.) to test
cp1-req create-database <<EOF
{
    "id": "appdb",
    "spec": {
      "database_name": "appdb",
      "database_users": [
        {
          "username": "appdb_admin",
          "password": "password",
          "db_owner": true,
        }
      ],
      "port": 5432,
      "patroni_port": 8426,
      "nodes": [
        {
          "name": "n1",
          "host_ids": ["host-1"]
        },
        {
          "name": "n2",
          "host_ids": ["host-1"]
        }
      ]
    }
  }
EOF

# You should get an error about duplicate ports
```

## Notes for Reviewers

The ticket mentions systemd, but the same issue affected Swarm as well. 

PLAT-611

### DIFF
--- a/changes/unreleased/Fixed-20260616-150152.yaml
+++ b/changes/unreleased/Fixed-20260616-150152.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: Database specs with duplicate allocated ports on a single host are now rejected by the API.
+time: 2026-06-16T15:01:52.449574-04:00
--- a/server/internal/api/apiv1/errors.go
+++ b/server/internal/api/apiv1/errors.go
@@ -10,6 +10,7 @@ import (
 	"github.com/pgEdge/control-plane/server/internal/database"
 	"github.com/pgEdge/control-plane/server/internal/etcd"
 	"github.com/pgEdge/control-plane/server/internal/task"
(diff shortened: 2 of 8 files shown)


moizpgedge (approved):
Looks good and it fixes the bug. I tested on my dev env: two nodes on the same host with the same ports got rejected, and same ports on different hosts still worked.

PR #411, opened 2026-06-17
feat: validate conf libraries contain spock

GabriellePoncey:
## Summary

Prevent users from disabling the spock extension via postgresql_conf.shared_preload_libraries. The control plane has a hard dependency on spock, so removing it would silently break replication.

## Changes

- Add validateConfLibraries to reject any postgresql_conf that explicitly sets shared_preload_libraries without including spock

- Wire into validateDatabaseSpec (spec-level conf) and validateNode (node-level conf) for the create path

- Wire into validateDatabaseUpdate for the update path, at both spec and node level

## Testing

- TestValidateConfLibraries: unit tests covering absent key, valid configs, uppercase param name, and missing spock

- Invalid postgresql_conf cases added to TestValidateDatabaseSpec and TestValidateNode to confirm the check is called in the right places

- TestValidateDatabaseUpdate_LibraryConf: confirms the check runs on update requests at both spec and node level

### DIFF
--- a/server/internal/api/apiv1/validate.go
+++ b/server/internal/api/apiv1/validate.go
@@ -42,17 +42,24 @@ func validateAuthFileGUCs(conf map[string]any, path validation.Path) []error {
 }
 
 func validateConfLibraries(conf map[string]any, path validation.Path) []error {
-	val, ok := utils.TypedFromMap[string](conf, "shared_preload_libraries")
-	if !ok {
-		return nil // defaults will be applied, including spock
-	}
-	for _, lib := range strings.Split(val, ",") {
-		if strings.TrimSpace(lib) == "spock" {
-			return nil
+	for key, val := range conf {
+		if strings.ToLower(strings.TrimSpace(key)) != "shared_preload_libraries" {
+			continue
 		}
+		strVal, ok := val.(string)
+		if !ok {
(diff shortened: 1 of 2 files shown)


server/internal/api/apiv1/validate.go
```diff
@@ -41,6 +41,27 @@ func validateAuthFileGUCs(conf map[string]any, path validation.Path) []error {
 	return errs
 }
 
+func validateConfLibraries(conf map[string]any, path validation.Path) []error {
+	for key, val := range conf {
```
jason-lynch, line 46: Could you please add a comment here noting that we can't do a simple lookup because parameter names are case-insensitive? Your implementation is good, and I want to make sure we don't accidentally change it in the future.

jason-lynch (commented):
This implementation is looking good! Could you please also add some tests for this function to `server/internal/api/apiv1/validate_test.go`. I'm thinking maybe:
- Add a `TestValidateConfLibraries` to test this all the cases of this function individually
- Just to validate that it's being called in the right place, add an invalid `PostgresqlConf` and expected error message to the `invalid` test cases in both `TestValidateDatabaseSpec` and `TestValidateNode`

jason-lynch (approved):
Nice work!

PR #419, opened 2026-06-30
feat: propagate task cancellation to in-progress activities

GabriellePoncey:
When a task is cancelled, ApplyEvent now detects it via an etcd watch and cancels its own context, interrupting long-running operations rather than running to completion. Operations that respect context cancellation (eg Patroni HTTP calls) are interrupted promptly while others complete their current step before the activity exits.

## Changes
- Add `task.Watcher`: subscribes to a task key in etcd and closes a
  `Done` channel when the task reaches a terminal state or is deleted;
  `Err` returns the reason (`ErrTaskCanceled`, `ErrTaskFailed`, or nil).
- Add `task.Service.NewWatcher` and `task.TaskStore.Watch` to create
  watchers from the service layer.
- Add `watcherRegistry` to `task.Service`: concurrent `ApplyEvent` calls
  watching the same task share one etcd stream rather than each opening
  their own; events are fanned out to all subscribers.
- Wire the watcher into `ApplyEvent`: a goroutine cancels the activity
  context when `watcher.Done()` fires. Watcher setup failure is
  non-fatal — the activity continues without interruption support.
- Treat `StatusCanceling` the same as `StatusCanceled` so activities are
  interrupted during the canceling phase, not only after it completes.

PLAT-583

### DIFF
--- a/server/internal/task/service.go
+++ b/server/internal/task/service.go
@@ -12,15 +12,25 @@ import (
 	"github.com/pgEdge/control-plane/server/internal/storage"
 )
 
+// NewWatcher creates a Watcher that monitors the given task and closes its
+// Done channel when the task reaches a terminal state or is deleted. Multiple
+// callers watching the same task share a single etcd watch stream. The caller
+// must call Close on the returned Watcher when done with it.
+func (s *Service) NewWatcher(_ context.Context, scope Scope, entityID string, taskID uuid.UUID) (*Watcher, error) {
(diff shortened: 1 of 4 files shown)


moizpgedge:
Hi @GabriellePoncey can you give a look  at coderabbit issues

moizpgedge (approved):
Approving. Tested cancel end to end and it works, the in progress activity stops right after cancel instead of running to the end.

PR #420, opened 2026-07-02
perf: reduce E2E CI runtime via parallel jobs and test tuning

moizpgedge:
# [PLAT-660 ](https://pgedge.atlassian.net/browse/PLAT-660)PR Summary

## What this does

This PR makes the CI build much faster. It used to take about 44 minutes and now takes about 18 minutes. That is roughly 59 percent faster. The main idea was to run things at the same time instead of one after another, plus some smaller fixes to slow and flaky tests.

## What changed

1. The CI used to run everything in one big job. Now it runs in 3 separate jobs at the same time called Basic check 1, Basic check 2, and Basic check 3.

2. Basic check 3 (the e2e tests) was the slowest part, so that job itself is now split across 3 machines running at the same time instead of one machine running everything.

3. Some database tests were waiting on a slow setting called a checkpoint. Lowering that setting made one test almost 40 percent faster. We checked other tests too and found this trick only helps tests that add a replica to an already running database, not tests that create everything fresh, so we only applied it where it actually helps.

4. Two tests that were doing similar setup work were combined into one test with two steps, since the setup itself was not the important part being tested.

5. A few tests were flaky because they were not given enough time to finish under load. We gave them more time so they pass reliably instead of randomly failing and needing a retry.

6. We also fixed a small bug in the Makefile where certain test filters would break if they included special characters.

## How we tested it

We ran the full test suite locally against a real cluster and everything passed. We also triggered several real CI runs on this branch to confirm each change actually works, not just in theory. The final run showed all jobs passing in about 18 to 19 minutes.

## Things to know

The Basic check 3 split relies on a CircleCI feature that only works inside their actual system, so it could only be tested by running it for real, not locally. The balance between its 3 machines will get better automatically over the next few runs as the system learns how long each test actually takes. Right now if Basic check 1 fails, the other two heavier jobs still run instead of stopping early. This uses a bit more compute time but keeps the speed benefit of running everything at once.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,8 +61,8 @@ commands:
               docker-compose-plugin
             sudo systemctl start docker
 jobs:
-  # basic_check_1: unit tests + lint + license check (make ci)
-  basic_check_1:
+  # basic_check: unit tests + lint + license check (make ci)
+  basic_check:
     executor: common
     steps:
       - common_setup
@@ -72,8 +72,7 @@ jobs:
       - store_test_results:
           path: .
 
-  # basic_check_2: multi-host cluster integration tests (clustertest/)
-  basic_check_2:
+  test_cluster:
     executor: common
     steps:
       - common_setup
@@ -87,8 +86,7 @@ jobs:
       - store_test_results:
           path: .
 
-  # basic_check_3: end-to-end tests (e2e/), split across 3 parallel containers
-  basic_check_3:
+  test_e2e:
     executor: common
     parallelism: 3
     steps:
@@ -215,9 +213,9 @@ workflows:
   test:
     unless: << pipeline.parameters.build_ref >>
     jobs:
-      - basic_check_1
-      - basic_check_2
-      - basic_check_3
+      - basic_check
+      - test_cluster
+      - test_e2e
 
   release:
     jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,13 +61,20 @@ commands:
               docker-compose-plugin
             sudo systemctl start docker
 jobs:
-  test:
+  unit_lint:
     executor: common
     steps:
       - common_setup
       - run:
           name: Run CI checks
           command: make ci TEST_RERUN_FAILS=2
+      - store_test_results:
+          path: .
+
+  test_cluster:
+    executor: common
+    steps:
+      - common_setup
       - run:
           name: Run cluster tests
           command: |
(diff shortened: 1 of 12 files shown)


.circleci/config.yml
```diff
@@ -61,25 +61,68 @@ commands:
               docker-compose-plugin
             sudo systemctl start docker
 jobs:
-  test:
+  # basic_check_1: unit tests + lint + license check (make ci)
+  basic_check_1:
```
jason-lynch: I think the name `basic_check` without the `_1` makes sense here. What do you think?

.circleci/config.yml
```diff
@@ -61,25 +61,68 @@ commands:
...
     steps:
       - common_setup
       - run:
           name: Run CI checks
           command: make ci TEST_RERUN_FAILS=2
+      - store_test_results:
+          path: .
+
+  # basic_check_2: multi-host cluster integration tests (clustertest/)
+  basic_check_2:
```
jason-lynch: I thought your original name, `test_cluster` was appropriate here, since this is for the `make test-cluster` target.

.circleci/config.yml
```diff
@@ -61,25 +61,68 @@ commands:
...
           command: |
             make docker-swarm-init
             make start-local-registry
             make buildx-init
             make test-cluster CONTROL_PLANE_IMAGE_REPO=172.17.0.1:5000/control-plane TEST_RERUN_FAILS=2
+      - store_test_results:
+          path: .
+
+  # basic_check_3: end-to-end tests (e2e/), split across 3 parallel containers
+  basic_check_3:
```
jason-lynch: I think your original name, `test_e2e` was a better fit.

Makefile
```diff
@@ -124,6 +124,7 @@ test-e2e:
 	$(gotestsum) \
 		--format-hide-empty-pkg \
 		--format standard-verbose \
+		--junitfile test-e2e-results.xml \
```
jason-lynch: I don't think this change is necessary. We automatically add the `--junitfile` argument on line 62 where we set `gotestsum`. If you look in the log output from the latest CircleCI run, you'll see that this argument is getting set twice now:

```sh
/home/circleci/.go_workspace/bin/gotestsum --junitfile test-e2e-results.xml \
        --format-hide-empty-pkg \
        --format standard-verbose \
        --junitfile test-e2e-results.xml \
        --rerun-fails=2 \
        --rerun-fails-max-failures=4 \
        --packages='./e2e/...' \
        -- \
        -tags=e2e_test -count=1 -timeout=45m -parallel 4 -run "^(TestAddNodeOriginAdvanced|TestCancelDatabaseTask|TestUpdateAddNode23|TestUpdateAddNode12|TestMinorVersionUpgrade|TestPortChange|TestScripts|TestWholeCluster)$" -args -fixture ci  -debug
```

jason-lynch (commented):
Nice work! This looks like a huge improvement. Sorry for the confusion around the job names - I was only suggesting `basic_check` as an alternative for the `make ci` job because it's doing several different types of tests and checks. I also have one very small suggestion in the Makefile, but otherwise, this looks really great!

.circleci/config.yml
moizpgedge: sure i can change it

.circleci/config.yml
moizpgedge: Changed this to basic_check 
because other are renamed.

jason-lynch (approved):

PR #423, opened 2026-07-02
chore: support IPv6 fallback in IP autodetection

GabriellePoncey:
## Summary

If no IPv4 address is found, fall back to the first available IPv6 address rather than returning an error.

## Changes

- Updated getFirstIP to fall back to the first routable IPv6 address if no IPv4 address is found
- Added GetBindAddr() utility that scans interfaces and returns the appropriate wildcard (:: for IPv6/dual-stack, 0.0.0.0 for IPv4-only)
- HTTP server address formatting now uses net.JoinHostPort, which correctly brackets IPv6 addresses in host:port strings
- etcd embedConfig uses GetBindAddr() to select a single wildcard listener, avoiding EADDRINUSE from conflicting dual-stack socket binds

## Testing

- Manual verification on a host with only IPv6 interfaces

### DIFF
--- a/server/internal/etcd/embedded.go
+++ b/server/internal/etcd/embedded.go
@@ -669,14 +669,12 @@ func embedConfig(cfg config.Config, logger zerolog.Logger) (*embed.Config, error
 
 	clientPort := cfg.EtcdServer.ClientPort
 	peerPort := cfg.EtcdServer.PeerPort
-	// Use :: on IPv6/dual-stack hosts (dual-stack socket via IPV6_V6ONLY=0 on Linux).
-	// Fall back to 0.0.0.0 on IPv4-only hosts where :: may not be available.
-	wildcard := "0.0.0.0"
-	if len(cfg.PeerAddresses) > 0 {
-		if ip := net.ParseIP(cfg.PeerAddresses[0]); ip != nil && ip.To4() == nil {
-			wildcard = "::"
-		}
+
+	wildcard, err := utils.GetBindAddr()
+	if err != nil {
+		return nil, fmt.Errorf("failed to detect bind address: %w", err)
 	}
+
 	c.ListenClientUrls = []url.URL{
 		{Scheme: "https", Host: net.JoinHostPort(wildcard, strconv.Itoa(clientPort))},
 	}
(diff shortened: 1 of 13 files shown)


server/internal/etcd/embedded.go
```diff
@@ -669,13 +669,19 @@ func embedConfig(cfg config.Config, logger zerolog.Logger) (*embed.Config, error
 
 	clientPort := cfg.EtcdServer.ClientPort
 	peerPort := cfg.EtcdServer.PeerPort
+	// Use :: on IPv6/dual-stack hosts (dual-stack socket via IPV6_V6ONLY=0 on Linux).
```
jason-lynch: Can we please revert this change for now? I understand that CodeRabbit flagged this as a problem, but this code has worked for a long time without issues, so I'd like to understand the conditions under which it breaks before we change it. This is something we can test out.

jason-lynch (approved):
Nice work! My one comment is not a functional change, so feel free to take it or leave it and merge whenever you're happy with it.

server/internal/utils/utils.go
```diff
@@ -155,6 +156,43 @@ func BuildOptionArgs(options map[string]string) []string {
...
+			switch v := addr.(type) {
+			case *net.IPNet:
+				ip = v.IP
+			case *net.IPAddr:
+				ip = v.IP
+			}
+			if ip == nil || ip.IsLoopback() || ip.IsLinkLocalUnicast() {
+				continue
+			}
+			if ip.To4() == nil {
```
jason-lynch: Just FYI: I've deleted a comment I had here. I had asked about inverting this condition and using `To16` instead of `To4` in `GetBindAddr`, but after [looking at the code](https://go.dev/src/net/ip.go), it turns out that `To16` returns both IPv4 and IPv6, so it's not the direct opposite of `To4`. `To4` also handles a special case for IPv4s represented as IPv6. After thinking about this some more, it might be clearest to do:

```suggestion
			if ip.To4() == nil && ip.To16() != nil {
```

It's probably not possible today for this IP to be neither an IPv4 nor an IPv6, so this is just readability feedback, not a blocking change.

PR #428, opened 2026-07-06
feat: change password_encryption default to scram-sha-256

moizpgedge:
# [PLAT-626](https://pgedge.atlassian.net/browse/PLAT-626)  Summary

## What this does

This PR changes the default Postgres password setting from `md5` to
`scram-sha-256`. Postgres itself has used `scram-sha-256` as its default
since version 16, but our control plane kept the older `md5` default
carried over from Cloud. `scram-sha-256` is the more secure option, and
`md5` password support is deprecated upstream and will eventually be
removed from Postgres entirely.

## What changed

1. The default value for `password_encryption` in
   `server/internal/postgres/gucs.go` was changed from `md5` to
   `scram-sha-256`. This is a one-line change. Everything downstream
   (the `pg_hba.conf` authentication rule, the golden test files) already
   reads this value dynamically, so nothing else needed to change to make
   it take effect.

2. No new code was needed to make existing databases pick up the new
   default automatically. The control plane already re-applies every
   `database_users` password via `ALTER ROLE ... WITH PASSWORD` any time
   a database is updated, regardless of whether the password itself
   changed. Since Postgres hashes a password using whatever
   `password_encryption` is active at that moment, this means every user
   listed in `database_users` gets rehashed to `scram-sha-256`
   automatically the next time that database goes through an update. No
   migration step is required. This also works even if the update
   request omits the password field entirely, since the API layer
   already merges in the previously stored password before re-applying
   it.

3. This is a breaking change for one specific case: any Postgres role
   created outside of `database_users` (for example, by connecting
   directly and running `CREATE ROLE`) is never touched by this
   mechanism. Those passwords will stay on `md5` until someone resets
   them by hand. This should be called out in the release notes along
   with instructions for resetting those passwords.

4. Added a unit test (`TestDefaultGUCs`) asserting the new default, and
   regenerated the six golden test files under
   `server/internal/orchestrator/common/golden_test/` that previously
   captured the old `md5` value in both the GUC list and the `pg_hba.conf`
   catch-all rule.

5. Added `e2e/password_encryption_test.go` with two new tests. One
   confirms a brand new database defaults straight to `scram-sha-256`
   with no configuration. The other confirms a database that starts on
   `md5` gets every `database_users` password automatically rehashed to
   `scram-sha-256` after an update, without resubmitting any password.

## How we tested it

We ran the full unit test suite, the specific golden tests, and `go test
-race` across the affected packages, and everything passed. We also ran
a long manual verification pass against a real local dev cluster, not
just unit tests. This covered a fresh database with no override
defaulting to `scram-sha-256`, and an existing `md5` database
automatically rehashing every `database_users` password after an update.

We also confirmed the rehash still works even when the update request
leaves the password field out entirely, which is the realistic way most
clients will actually call this. Beyond that we tested a three node
distributed database, where every node independently ends up on
`scram-sha-256` and all of them authenticate correctly, and adding a
fourth node to that cluster, which uses a different code path (copying
an existing node's already hashed role dump) and still works.

Other cases we checked: password rotation, a user with no password at
all, and downgrading back to `md5` explicitly to confirm the mechanism
works in both directions. We also created a role outside of
`database_users` and confirmed its password stayed byte for byte
unchanged across two separate updates, which is exactly the breaking
change behavior we want. An invalid `password_encryption` value is
safely rejected by Postgres at reload time with no downtime. Finally we
ran a full pgBackRest backup and restore cycle and confirmed restored
data comes back correctly and the restored role gets a fresh
`scram-sha-256` hash.

[PLAT-626]: https://pgedge.atlassian.net/browse/PLAT-626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


e2e/password_encryption_test.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/jackc/pgx/v5"
+	"github.com/stretchr/testify/require"
+
+	api "github.com/pgEdge/control-plane/api/apiv1/gen/control_plane"
+)
+
+// TestPasswordEncryptionDefaultsToScram checks that a newly created database,
+// which does not override password_encryption, hashes database_users
+// passwords with scram-sha-256 rather than the legacy md5 default.
+func TestPasswordEncryptionDefaultsToScram(t *testing.T) {
```
jason-lynch: I don't think we need a dedicated E2E for this. The golden tests are sufficient for enforcing our database configuration, and it looks like you added a unit test as well. Each E2E test that creates a database takes a minute or two to execute, so we need to be judicious about when that cost is justified.

jason-lynch (commented):
Nice! I left one minor suggestion regarding E2Es. Could you also please include a changelog entry for this? The important things to note in that entry are:
- We've changed the default password encryption
- If you're using the default password encryption, all roles in `database_users` will be automatically updated the next time you update
- Any roles that were created outside of `database_user` will need their passwords updated manually
- You can keep using `md5` password encryption by setting `postgresql_conf.password_encryption` in your database spec

e2e/password_encryption_test.go
moizpgedge: Done

jason-lynch (approved):

PR #435, opened 2026-07-16
fix: patch CVEs flagged by Trivy image scan

moizpgedge:
## Summary

Fixes the Trivy image scan findings (2 Critical, 28 High, 17 Medium, 1 Low) by bumping the vulnerable deps and Go toolchain, risk-accepting the couple of CVEs with no fix available, hardening the Dockerfiles, and fixing the licenses-ci CI failure that came out of it.

## Changes

**Dependency bumps.** Bumped pgx to v5.10.0 (fixes two Critical CVEs), opentelemetry to v1.44.0, x/crypto to v0.54.0, x/net to v0.57.0, and the Go toolchain itself to 1.26.5, which alone covers 11 stdlib CVEs. These were all just following the Trivy report's fixed-version column.

**Risk acceptances.** docker/docker's CVEs don't have a fix in the v27 line we're on — confirmed with govulncheck that every one says "Fixed in: N/A" — and we don't use any of the archive/copy APIs the vulnerable code paths are in anyway. x/crypto/openpgp's advisory doesn't apply since we don't import it. Both documented in trivyignore.

**Dockerfile hardening.** Pinned the FROM images to digests instead of floating latest tags, added --no-install-recommends to the CI Dockerfile. All four Dockerfiles still run as root on purpose — the binary needs to chown bind-mounted Postgres/Patroni data for arbitrary DB-owner UIDs, and distroless has no shell to drop privileges with anyway. docs/Dockerfile has its own reason: the doc build bind-mounts the whole repo with no UID mapping, so a fixed non-root user could just as easily break on a different host UID. Tried the inline trivy:ignore comment for this first but it doesn't actually work (it's a whole-file check, not line-scoped, confirmed by testing it), so moved the suppression to a plain .trivyignore entry with the reasoning written down instead.

**NOTICE.txt fix.** go-licenses has a bug where it generates duplicate-path LICENSE links for some opentelemetry-go packages, giving 404s. Fixed it in the Makefile itself (a sed step, not a one-off edit) so regeneration stays correct going forward, then regenerated NOTICE.txt for real in a container matching CI's Go version — my local macOS Go can't even run go-licenses. This is also what fixes the licenses-ci failure, since an earlier manual edit to NOTICE.txt couldn't survive CI regenerating and diffing it.

## Testing

go build, go vet clean. go test passes except one pre-existing unrelated failure in the mqtt package that needs a rootless-docker socket, nothing to do with this PR. Cross-checked the docker/docker CVEs with govulncheck. Rebuilt the actual production image and rescanned it with Trivy: zero vulnerabilities at any severity, and confirmed with --show-suppressed that only the 6 documented docker/docker CVEs are hidden, nothing else. Rescanned all four Dockerfiles for misconfigs: clean except the pre-existing missing-HEALTHCHECK low finding, out of scope here. Also checked for secrets and license issues on the built image, nothing concerning. Ran make licenses in a matching container and confirmed zero diff against the committed NOTICE.txt.

## Checklist

- [x] Tests — N/A, this is dependency/config only, verified via the rescans above
- [x] Documentation — NOTICE.txt regenerated, risk acceptances and root-user reasoning documented inline
- [x] Issue linked — PLAT-686
- [x] Changelog entry added
- [ ] No breaking changes


docker/control-plane/Dockerfile
```diff
@@ -1,10 +1,17 @@
-FROM gcr.io/distroless/static-debian12
+FROM gcr.io/distroless/static-debian12:latest@sha256:61b7ccecebc7c474a531717de80a94709d20547cdcdaf740c25876f2a8e38b44
```
tsivaprasad, line 1: Is this digest a fat manifest (multi-arch) or platform-specific?

moizpgedge: Yep, this image works on all the different chip types (Intel/AMD and ARM included), so it's fine

tsivaprasad (approved):

PR #436, opened 2026-07-17
fix: block SQL injection in role attributes

moizpgedge:
## Summary

Fixes a real SQL injection bug in role creation, cleans up a false-positive finding on the same scan, risk-accepts a couple of CVEs that don't have fixes yet, and fixes the licenses-ci CI failure.

## Changes

**SQL injection fix.** `CreateUserRole` in `roles.go` was building `ALTER ROLE ... WITH <attr>` by pasting each attribute string straight into the SQL. A bad attribute like `"LOGIN; DROP TABLE users; --"` would just run. Added an allowlist of the actual valid role attribute keywords (SUPERUSER, CREATEDB, LOGIN, etc. plus their NO* forms) and reject anything else before it gets near a SQL statement. Left out the attribute forms that take a literal value (CONNECTION LIMIT, VALID UNTIL, PASSWORD) since those can't be safely allowlisted this way and password already has its own field. Added tests in roles_test.go. Also tried this for real against a running dev cluster: sent the malicious attribute through the actual API, and it got rejected with "unsupported role attribute" before touching Postgres.

**False positive cleanup.** The scanner also flagged `PostgresContainerExec` in utils.go as SQL injection, but it's a Docker exec call, not SQL. Checked every place that calls it and confirmed the command is always built from fixed binaries, never raw input, and Docker doesn't run it through a shell anyway. Added a comment explaining why so it doesn't keep getting flagged.

**Risk acceptances.** Added two entries to the trivyignore file: the docker/docker CVEs that have no fix in the v27 line we're on (and we don't use the code paths they affect anyway), and the x/crypto/openpgp advisory, which doesn't apply since we don't import that package.

**Fixed the licenses-ci failure.** NOTICE.txt had gotten out of sync with go.mod, and separately go-licenses has a bug where it generates broken LICENSE links for some opentelemetry-go packages. Fixed the link bug in the Makefile so it's part of regeneration going forward instead of a one-off patch, then regenerated NOTICE.txt properly in a container that matches CI's Go version (couldn't do it locally on macOS, go-licenses crashes there for an unrelated reason). Ran it twice to make sure it's stable.

## Testing

go build, go vet, go test all pass. Tested the injection fix live against a dev cluster. Regenerated NOTICE.txt in a matching CI container twice, same output both times.

## Checklist

- [x] Tests added (roles_test.go)
- [x] Documentation updated (trivyignore, inline comments)
- [x] Issue linked (PLAT-685)
- [ ] Changelog entry — still need to add one
- [ ] No breaking changes, aside from attributes outside the allowlist now getting rejected instead of silently going through (nothing in this codebase relies on that)

## Notes for reviewers

The NOTICE.txt/Makefile fix also applies to the PLAT-686 branch, which hit the same CI failure for the same reason.


tsivaprasad (approved):
Looks good.

PR #437, opened 2026-07-20
docs: systemd docs recommend enterprise repo download

GabriellePoncey:
## Summary

 Recommend installing and updating the Control Plane package on systemd hosts from the pgEdge Enterprise dnf/apt repositories,since the control-plane package will begin publishing there starting with the next release. The existing GitHub Releases curl-based method is kept as a fallback for air-gapped hosts or pinning an exact version.

## Changes

 - Installing the Control Plane doc now recommends dnf install/apt install pgedge-control-plane from the enterprise repo, keeps original logic as a fallback subsection.
  - docs/installation/systemd-upgrading.md: same treatment for upgrades, enterprise repo recommended but fallback kept


GabriellePoncey:
@coderabbitai review

moizpgedge (approved):

PR #444, opened 2026-07-24
build(release): bump version to v0.10.0

tsivaprasad:
## Summary
This is the release PR for v0.10.0.

## Testing
Verification:

- All e2e tests passed on the Lima fixture.


.goreleaser.yaml
```diff
@@ -64,17 +64,57 @@ nfpms:
...
+        file_info:
+          mode: 0644
+      # Docs and license, at the same paths for both formats. The 'doc' and
+      # 'license' types add the %doc/%license markers to the rpm spec, but nfpm
+      # drops content of those types from non-rpm packages, so the deb needs its
+      # own plain-file entries (scoped with `packager` so each format gets
+      # exactly one entry per path).
+      - src: README.md
+        dst: /usr/share/doc/pgedge-control-plane/README.md
+        type: doc
```
jason-lynch, line 99: `doc` is not a valid `type` value. This field should be omitted.

```suggestion
```

.goreleaser.yaml
```diff
@@ -64,17 +64,57 @@ nfpms:
...
+        file_info:
+          mode: 0644
+      - src: README.md
+        dst: /usr/share/doc/pgedge-control-plane/README.md
+        packager: deb
+        file_info:
+          mode: 0644
+      - src: LICENSE.md
+        dst: /usr/share/licenses/pgedge-control-plane/LICENSE.md
+        type: license
```
jason-lynch, line 110: `license` is not a valid type value. This field should be omitted.

```suggestion
```

.goreleaser.yaml
```diff
@@ -64,17 +64,57 @@ nfpms:
...
+          mode: 0644
       - src: ".sbom/sbom.{{ .Format }}.asc"
         dst: /usr/share/pgedge-control-plane/pgedge-control-plane-sbom.json.asc
+        file_info:
+          mode: 0644
+      # Docs and license, at the same paths for both formats. The 'doc' and
+      # 'license' types add the %doc/%license markers to the rpm spec, but nfpm
+      # drops content of those types from non-rpm packages, so the deb needs its
+      # own plain-file entries (scoped with `packager` so each format gets
+      # exactly one entry per path).
```
jason-lynch: I don't think this comment is correct. As CodeRabbit pointed out, Debian and RHEL have different conventional locations for licenses, and Goreleaser's nFPM docs don't mention `doc` or `license` and those values are not defined in their published JSON schema: https://goreleaser.com/customization/package/nfpm/

```suggestion
```

.goreleaser.yaml
```diff
@@ -64,17 +64,57 @@ nfpms:
...
       - src: packaging/pgedge-control-plane.service
         dst: /usr/lib/systemd/system/pgedge-control-plane.service
         type: config
+        file_info:
+          mode: 0644
       - src: packaging/config.json
         dst: /etc/pgedge-control-plane/config.json
         type: config|noreplace
+        file_info:
+          mode: 0640
```
jason-lynch: I think `0600` would make more sense for this file than `0640`.

```suggestion
          mode: 0600
```

.goreleaser.yaml
```diff
@@ -64,17 +64,57 @@ nfpms:
         scripts:
           postinstall: packaging/deb/postinstall.sh
           preremove: packaging/deb/preremove.sh
+    # Every entry sets file_info.mode explicitly. Without it nfpm packages the
+    # mode of the file as checked out on the builder, which is umask-dependent
+    # (CI runs with umask 002, so git creates 0664 and that mode shipped in the
+    # packages). config.json is 0640 because it may hold credentials; the unit
+    # runs as root, so root-only read access is sufficient.
```
jason-lynch: I don't know how accurate this comment is. Having no comment is better than having an inaccurate comment, so let's remove it for now.

```suggestion
```

tsivaprasad: Updated as per https://github.com/goreleaser/nfpm/blob/v2.44.1/nfpm.go#L586-L587  & https://github.com/goreleaser/nfpm/blob/v2.44.1/files/files.go#L152-L155

.goreleaser.yaml
tsivaprasad: Looks like the docs just haven't been updated — the actual `nfpm` source (`files.go, rpm.go`) defines and handles both doc and license explicitly. https://github.com/goreleaser/nfpm/blob/v2.44.1/files/files.go#L44-L49

.goreleaser.yaml
tsivaprasad: here also valid it seems https://github.com/goreleaser/nfpm/blob/main/files/files.go#L57

.goreleaser.yaml
tsivaprasad: Seems its valid `type` https://github.com/goreleaser/nfpm/blob/v2.44.1/files/files.go#L45

.goreleaser.yaml
jason-lynch: Thanks for tracking that down! But the first part of the comment is now incorrect with your other changes:

> Docs and license, at the same paths for both formats.

It's not a big deal. We can just fix it in main and keep going with this release.

jason-lynch (dismissed):

jason-lynch (approved):

PR #448, opened 2026-07-30
docs: simplify systemd install instructions

jason-lynch:
## Summary

This commit condenses the dependency and pgedge-control-plane package installation instructions and removes the manual package installation instructions. We depend on packages from the pgEdge Enterprise Postgres repository, so installing the packages manually isn't a true alternative to our package repos.


tsivaprasad (approved):
@jason-lynch Have you run this changes as per https://github.com/pgEdge/pgedge-skills? otherwise looks good

PR #450, opened 2026-07-30
docs: Restish cli guide

GabriellePoncey:
## Summary

Adds a guide for using restish as a cli tool. Document a quickstart process (rapid installation and experimentation), profile management, project configuration and mtls profile setup, database config files and secrets.

PLAT-671

### DIFF
--- a/docs/api/restish.md
+++ b/docs/api/restish.md
@@ -217,7 +217,10 @@ time, though, so pass that from a separate file you don't commit instead
 of adding it to `databases/example.json`:
 
 ```sh
-cat > /tmp/example.create.json <<'EOF'
+umask 077
+tmpfile=$(mktemp)
+trap 'rm -f "$tmpfile"' EXIT
+cat > "$tmpfile" <<'EOF'
 {
     "id": "example",
     "spec": {
@@ -239,9 +242,14 @@ cat > /tmp/example.create.json <<'EOF'
     }
 }
 EOF
-restish pgedge create-database < /tmp/example.create.json
-rm /tmp/example.create.json
+restish pgedge create-database < "$tmpfile"
 ```
+
+`umask 077` keeps the file unreadable by anyone else on the machine while
+it exists, `mktemp` avoids a predictable filename, and the `trap` removes
+it as soon as this shell exits — including if `create-database` itself
+fails — rather than relying on a final `rm` that a failure or an
+interrupted copy-paste could skip.
 
 Update the same database by editing `databases/example.json` and
 re-applying it against the `update-database` command. No password is
@@ -278,11 +286,13 @@ already stored unless you explicitly send a new one. See
 In practice, that means the default workflow is the one shown above: keep
 `databases/example.json` secret-free from the start, and pass real secret
 values only from a separate, uncommitted file for the one `create-database`
-call that needs them — deleting that file immediately afterward rather than
-editing secrets out of the committed file after the fact. From then on,
-`update-database` runs against the secret-free file as-is. If you need to
-rotate a password, apply it the same way: a temporary file with the new
-value, passed once, then removed.
+call that needs them — a file created with a restrictive `umask`, an
+unpredictable `mktemp` path, and an `EXIT` trap so it's removed even if
+the command fails, rather than editing secrets out of the committed file
+after the fact. From then on, `update-database` runs against the
+secret-free file as-is. If you need to rotate a password, apply it the
+same way: the same `umask`/`mktemp`/`trap` pattern, with the new value,
+passed once.
 
 This keeps `databases/example.json` safe to read, diff, and share at any
 point — it's never the file that held the credential, so there's no window
--- a/docs/api/restish.md
+++ b/docs/api/restish.md
@@ -5,7 +5,7 @@ The pgEdge Control Plane makes use of a tool called [Restish](https://rest.sh) t
 Jump to what you're trying to do:
 
 - **[Quickstart](#quickstart)** — install Restish and run your first command against a cluster.
-- **[Managing Multiple Environments](#managing-multiple-environments)** — one cluster isn't enough, or you want a persistent setup you (and your team, if you have one) can reuse instead of reconnecting by hand every time.
+- **[Managing Multiple Environments](#managing-multiple-environments)** — one cluster isn't enough, or you want a persistent setup.
 - **[Managing Database Configuration as Files](#managing-database-configuration-as-files)** — commit database specs to source control instead of typing JSON inline.
 
 ## Quickstart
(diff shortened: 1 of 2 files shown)


docs/api/restish.md
```diff
@@ -0,0 +1,293 @@
...
+
+### 1. Installation
+
+First, install Restish [via Restish's official website](https://rest.sh/docs/getting-started/install/). Restish supports many installation methods, including Homebrew (macOS), GitHub Releases, and OCI images; select the option that most aligns with your organization's preferences and practices.
+
+### 2. Connection
+!!! warning
+
+    Only connect this way to clusters and databases you're okay with experimenting on. See [Managing Multiple Environments](#managing-multiple-environments)
+    before connecting Restish to anything production.
```
jason-lynch, line 26: I don't know if this needs to be a warning. Maybe just a note to the effect of "see this section if you need to connect to more than one environment"? The warning makes it sound like Restish itself is unsafe for production.

docs/api/restish.md
```diff
@@ -0,0 +1,293 @@
...
+[installation quickstart](../installation/quickstart.md) if not), you can
+be running commands against it in three steps:
+
+
+### 1. Installation
+
+First, install Restish [via Restish's official website](https://rest.sh/docs/getting-started/install/). Restish supports many installation methods, including Homebrew (macOS), GitHub Releases, and OCI images; select the option that most aligns with your organization's preferences and practices.
+
+### 2. Connection
+!!! warning
```
jason-lynch, line 23: Very minor - there should probably be a newline here.

```suggestion

!!! warning
```

docs/api/restish.md
```diff
@@ -0,0 +1,293 @@
...
+generated command, and once you have a database config file (see
+[Managing Database Configuration as Files](#managing-database-configuration-as-files)
+below), `restish pgedge create-database < your-file.json` creates one.
+
+## Managing Multiple Environments
+
+Restish doesn't enforce any naming convention for the APIs you connect to.
+We recommend using Restish's **profiles** feature: one API registration, `pgedge`, holds a profile per environment, and each profile can override the base URL (and, if you need it later, auth or other per-environment request details).
+
+**Use descriptive cluster names.** Every cluster has its own durable `id`,
```
jason-lynch, line 52: I feel like we should mention here that the ID cannot be changed after initialization. Maybe something like:

```suggestion
**Use descriptive cluster names.** Every cluster has an immutable `id`,
```

docs/api/restish.md
```diff
@@ -0,0 +1,293 @@
...
+set at initialization and returned by `get-cluster`. `init-cluster` takes
+an optional `cluster_id` query parameter. Setting it to something descriptive will allow you to keep track of multiple different clusters.
+
+```sh
+curl "http://host1.internal:3000/v1/cluster/init?cluster_id=production"
```
jason-lynch, line 57: Could you please change this to the `restish` equivalent? Or maybe use tabs to list both, with the restish tab as the default?

docs/api/restish.md
```diff
@@ -0,0 +1,293 @@
...
+
+!!! note
+
+    A project config only honors the `apis` and `theme` keys — nothing
+    else. Keep it secret-free: the Control Plane's dev and local endpoints
+    don't need auth, so these profiles can stay limited to `base_url` as
+    shown above. If a cluster you connect to does require auth, reference
+    the value as `env:NAME` rather than committing it literally.
+
+If a cluster has [mTLS enabled](../installation/mtls.md), add the CA and
```
jason-lynch, line 136: I think these `mTLS` instructions should go in their own section.

docs/api/restish.md
```diff
@@ -0,0 +1,293 @@
...
+
+    A cluster is made up of multiple hosts, each with its own host ID
+    (e.g. `host-1`). Any of them can serve a request for the whole
+    cluster, which is why one `base_url` per environment is enough for
+    routine use — you don't need a profile per host. To target a specific
+    host directly (retrying against a different one, or comparing behavior
+    across hosts while debugging), give it its own profile the same way:
+    `-p production-host-1`.
+
+### Persisting Connections in a Project Config
```
jason-lynch, line 85: I don't think we need to cover project configs in our doc. This setup will be atypical for most of our users, and we can just provide a link back to the restish docs for people who want to learn about the restish features we don't cover.

docs/api/restish.md
```diff
@@ -0,0 +1,293 @@
...
+
+Update the same database by editing `databases/example.json` and
+re-applying it against the `update-database` command. No password is
+needed, since it's omitted from the request entirely:
+
+```sh
+restish pgedge update-database example < databases/example.json
+```
+
+!!! note
```
jason-lynch, line 246: I don't think we need this note. We should never double-process a request unless there's a bug. We have multiple mechanisms in place to prevent it. In our development docs, we recommend disabling this retry feature entirely because 500s are common during active development, and the rapid retries make debugging harder. The retries might be similarly annoying for end users, but it shouldn't be dangerous unless multiple safeguards have failed.

docs/api/restish.md
```diff
@@ -0,0 +1,293 @@
...
+
+```sh
+restish api list                             # every connection you've configured
+restish api inspect pgedge-sandbox           # the URL, profiles, and spec Restish resolved
+restish api remove pgedge-sandbox            # disconnect (personal connections only)
+restish pgedge-sandbox --help                # every generated command
+restish pgedge-sandbox list-databases --help # options for one command
+```
+
+## Managing Database Configuration as Files
```
jason-lynch, line 181: It might be worth mentioning in this section that the Control Plane API only ever returns sanitized database specs without any secret values. Users who have many database users or other secrets in their spec might find it easier to use the API to ensure they have a sanitized spec for source control:

1. Create a database using a file or other request that contains your secret values
2. Use the `get-database` command to get the sanitized spec and store it in a file:

```sh
restish pgedge get-database example | jq '{ spec }' > databases/example.json
```

docs/api/restish.md
```diff
@@ -0,0 +1,293 @@
...
+restish -p production pgedge update-database example < databases/example.json
+```
+
+This gives you a directory of database configuration files you can commit
+to source control, diff, review in a pull request, and re-apply — the same
+workflow you'd use for any other infrastructure-as-code.
+
+### Handling Secrets
+
+The one field in these files that doesn't belong in source control is
```
jason-lynch, line 271: I think we have at least 15 sensitive fields (probably more) in our spec. Rather than giving a count, maybe we could just say something like, "Secret fields, such as `database_users[].password` or `s3_key_secret`, should be excluded from any files committed to source control."

jason-lynch (commented):
Awesome! This is looking good so far. I've left a couple of suggestions. Feel free to push back on any that you disagree with.

docs/api/restish.md
jason-lynch: It looks like you changed this to a note, but it's not formatted as an admonition (`!!! note`) and the wording still implies that this is unsafe. It's reasonable to assume that users will install restish on one or more of the control plane hosts and connect to it locally, so I don't think this is the right framing. I was trying to suggest something like:

```md
!!! note

    See [Managing Multiple Environments](#managing-multiple-environments)
    if you need to connect to multiple Control Plane clusters from one machine.
```

docs/api/restish.md
jason-lynch: It looks like this is still an issue.

docs/api/restish.md
```diff
@@ -0,0 +1,289 @@
...
+    The Control Plane excludes every secret field from its responses, you can skip manual redaction entirely: create the database from a one-off request that includes all of its secrets, then pull the sanitized spec back into your file:
+
+    ```bash
+    (
+      set -e
+      set -o pipefail
+      tmpfile=$(mktemp)
+      trap 'rm -f "$tmpfile"' EXIT
+      restish pgedge get-database example | jq '{ id, spec }' > "$tmpfile"
+      jq empty "$tmpfile"
```
jason-lynch: Is this `jq empty` line doing anything? It looks like the `empty` filter's purpose is to just output nothing.

jason-lynch (commented):
Thanks for making those changes! Still a few small issues to resolve.

PR #456, opened 2026-08-17
feat: add dedicated Spock major-version upgrade

moizpgedge:
Upgrading a database from Spock 5.x to 6.x requires rolling each node's binary forward one at a time rather than redeploying every node at once, since a fresh Spock 6 subscription cannot sync from a Spock 5.x peer (spock.read_peer_progress doesn't exist pre-6, verified empirically against a real mixed 5/6 cluster). Spock's own rolling- upgrade compatibility already covers the mixed-version window, confirmed both by their t/014_rolling_upgrade.pl TAP test and by a live end-to-end run of this workflow against a 6-host dev cluster.

The existing patch/minor upgrade path was never designed for this and had a real, previously-unenforced gap: ValidateChangedSpec checked Postgres major version changes but not Spock major version changes, so an ordinary update-database call could in theory bump every node's Spock major at once, which is exactly the scenario a rolling upgrade needs to avoid.

- Add a per-node spock_version override to the spec model, gated by server-side validation so it can only be set by the new workflow, never through create-database or update-database.
- Add spockMajorVersionChanged() to ValidateChangedSpec, closing the gap above.
- Add a dedicated apply-major-upgrade API action and MajorVersionUpgrade workflow that rolls the spec forward one node at a time, verifying each node is replicating before moving to the next, then normalizes the spec once every node is on the target version.
- Add FindMajorUpgrade to the orchestrator interface (Docker Swarm implementation; systemd returns a clear unsupported error).

[PLAT-720](https://pgedge.atlassian.net/browse/PLAT-720)

## Summary

Adds a dedicated `apply-major-upgrade` action and `MajorVersionUpgrade` workflow for rolling a database from Spock 5.x to 6.x one node at a time, kept fully independent of the existing patch/minor `apply-upgrade` path.

Upgrading from Spock 5.x to 6.x requires rolling each node's binary forward one at a time rather than redeploying every node at once, since a fresh Spock 6 subscription cannot sync from a Spock 5.x peer (`spock.read_peer_progress` doesn't exist pre-6, verified empirically against a real mixed 5/6 cluster). Spock's own rolling-upgrade compatibility already covers the mixed-version window, confirmed both by their `t/014_rolling_upgrade.pl` TAP test and by a live end-to-end run of this workflow against a 6-host dev cluster.

The existing patch/minor upgrade path was never designed for this and had a real, previously-unenforced gap: `ValidateChangedSpec` checked Postgres major version changes but not Spock major version changes, so an ordinary `update-database` call could in theory bump every node's Spock major at once — exactly the scenario a rolling upgrade needs to avoid.

## Changes

- Add a per-node `spock_version` override to the spec model, gated by server-side validation so it can only be set by the new workflow, never through `create-database` or `update-database`.
- Add `spockMajorVersionChanged()` to `ValidateChangedSpec`, closing the gap above.
- Add a dedicated `apply-major-upgrade` API action and `MajorVersionUpgrade` workflow that rolls the spec forward one node at a time, verifying each node is replicating before moving to the next, then normalizes the spec once every node is on the target version.
- Add `FindMajorUpgrade` to the orchestrator interface (Docker Swarm implementation; systemd returns a clear unsupported error).

## Testing

- `go build ./...`, `go vet ./...`, `golangci-lint run ./...` — all clean.
- `go test ./...` — full suite passes, including new unit tests:
  - `TestOrchestrator_FindMajorUpgrade` (6 subtests) covering image/version matching, cross-major-bucket rejection, and invalid input.
  - `TestService_ApplyMajorUpgrade` / `RollbackApplyMajorUpgrade` / `ApplyMajorUpgradeSpecChange` covering the happy path, not-modifiable rejection, `FindMajorUpgrade` failure propagation, `node_order` validation (missing/duplicate/unknown node), and rollback.
  - New `TestValidateChangedSpec` cases proving a Spock major change is rejected at both the database- and node-level via the ordinary `update-database` path — the acceptance criterion "existing minor/patch upgrade workflows cannot trigger cross-major upgrades."
  - New `TestValidateNode` case proving the per-node `spock_version` rejection fires on `create-database`/`update-database`.
- Manually verified end-to-end against a real 6-host Docker Swarm dev cluster: created a 3-node Spock 5.0.10 database with live bidirectional replication, ran the full rolling upgrade to Spock 6.0.0 (n1 → n2 → n3), and confirmed each already-upgraded node stayed `replicating` against its still-Spock-5 peers throughout — direct proof of the mixed 5.x/6.x compatibility finding this design rests on. Also verified live: data survives before/during/after the upgrade; the spec is correctly normalized afterward; a custom `node_order` is honored exactly; all three invalid `node_order` shapes are rejected before any node is touched; a concurrent racing upgrade request is rejected without disturbing the in-flight one; and both `apply-upgrade` and `update-database` refuse to cross the Spock major boundary with their designed error messages.

## Checklist

- [ ] Documentation updated (if needed) — `CLAUDE.md`'s workflow list is updated, and a private testing runbook was written and verified end-to-end, but the end-user `docs/using/upgrade-db.md` page (mkdocs site) does not yet mention this action. Flagging for discussion — see Notes for Reviewers.
- [x] Issue is linked (branch name and URL above)
- [x] [Changelog](https://github.com/pgEdge/control-plane/blob/main/docs/development/development.md#pull-requests) entry added for user-facing behavior changes
- [x] Breaking changes (if any) are clearly called out in the PR description — none; see Notes for Reviewers

## Notes for Reviewers

- **Not a breaking change, but a behavior tightening worth flagging**: an ordinary `update-database` call that bumps a database's Spock major version uniformly will now be rejected with `spock major version changed from X to Y; use the dedicated major-version upgrade action instead`. No legitimate workflow could have relied on the old behavior succeeding — `apply-upgrade`'s own `FindUpgrade` already independently rejected cross-Spock-major targets at the orchestrator layer — so this closes a latent gap rather than changing supported behavior.
- **End-user documentation is intentionally not included in this PR.** `docs/using/upgrade-db.md` already has a `## Major Version Upgrades` section, but it's about Postgres major version upgrades via the zero-downtime add-node approach — an unrelated, pre-existing mechanism. Adding Spock major-version upgrade docs there needs a naming decision to avoid confusion between "Postgres major upgrade" and "Spock major upgrade" in the same doc; happy to write it once we agree on the structure.
- **Compatibility finding**: confirmed via `git grep` across the full Spock 5.0.10 source tree (zero hits for `read_peer_progress`) and by live reproduction — a Spock 6 node's subscription to a Spock 5 peer fails outright with `function spock.read_peer_progress(unknown, integer, integer) does not exist`. This is why the design is a rolling in-place upgrade rather than an expand/contract approach.
- The dev-cluster testing runbook used for live verification lives outside this repo; happy to share it if useful for review, or fold relevant parts into `docs/development/` if there's interest in keeping it.

[PLAT-720]: https://pgedge.atlassian.net/browse/PLAT-720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jason-lynch (commented):
The normal 'update database' process already performs a rolling update, operating one node at a time and updating replicas before primaries. As you said in your description, we don't validate Spock version changes, so a user could use the normal 'update database' endpoint to change the Spock version in their spec, and the control plane will perform a rolling update.

Where/how does this new process differ from the existing update process?

moizpgedge:
@jason-lynch 
You're right that update database already does a rolling apply, replicas before primary, one pass across the spec. But there are two things it doesn't do that matter here.

First, it waits for the instance to come back up and Patroni to report healthy, but it never checks whether Spock replication actually resumed. It just assumes the extension's own upgrade handled it and moves on. That's fine for a minor version bump since the replication protocol doesn't change. For a major bump it's the actual risk: if a node's extension update doesn't cleanly bring replication back, update database has no way to notice and would just move on to the next node anyway. The new workflow checks that each node is actually replicating again before touching the next one.

Second, update database plans the whole spec change once at the start and then applies it. This workflow plans one node at a time, so each step is working off the cluster's real current state instead of a plan made before anything happened.

we don't currently block a Spock version change through update database, that's what spockMajorVersionChanged fixes here. But fixing that alone doesn't make update database safe for this, because of the first point. It just means it correctly says no now instead of doing something we can't verify.

So it's less that the rolling update is wrong and more that a Spock major bump needs a stronger per node guarantee than a normal spec edit does.

jason-lynch:
> First, it waits for the instance to come back up and Patroni to report healthy, but it never checks whether Spock replication actually resumed. It just assumes the extension's own upgrade handled it and moves on. That's fine for a minor version bump since the replication protocol doesn't change. For a major bump it's the actual risk: if a node's extension update doesn't cleanly bring replication back, update database has no way to notice and would just move on to the next node anyway. The new workflow checks that each node is actually replicating again before touching the next one.

That's a good point. We should discuss how we could incorporate this type of check into our existing process. For example, rather than basing it on the Spock version, we could raise an error if any existing Spock subscription goes from healthy to unhealthy.

> Second, update database plans the whole spec change once at the start and then applies it. This workflow plans one node at a time, so each step is working off the cluster's real current state instead of a plan made before anything happened.

You're right that the plan happens once, but we're always working off of the cluster's real state. We persist all resource updates back to the state as the update runs, so each resource always has access to the latest view of the database state. Planning each node individually will produce identical results, except it will duplicate some operations we would otherwise run only once.

> we don't currently block a Spock version change through update database, that's what spockMajorVersionChanged fixes here. But fixing that alone doesn't make update database safe for this, because of the first point. It just means it correctly says no now instead of doing something we can't verify.

This is a good change, but we should discuss what we should and shouldn't allow. For example, we might want to allow a major version update, but we should probably block downgrades.

In its current state, I'm not convinced we need a dedicated endpoint and so much new code for this process. I think we can do a much slimmer implementation that just adds the API validation and the replication status validation.

tsivaprasad:
> > First, it waits for the instance to come back up and Patroni to report healthy, but it never checks whether Spock replication actually resumed. It just assumes the extension's own upgrade handled it and moves on. That's fine for a minor version bump since the replication protocol doesn't change. For a major bump it's the actual risk: if a node's extension update doesn't cleanly bring replication back, update database has no way to notice and would just move on to the next node anyway. The new workflow checks that each node is actually replicating again before touching the next one.
> 
> That's a good point. We should discuss how we could incorporate this type of check into our existing process. For example, rather than basing it on the Spock version, we could raise an error if any existing Spock subscription goes from healthy to unhealthy.
> 
> > Second, update database plans the whole spec change once at the start and then applies it. This workflow plans one node at a time, so each step is working off the cluster's real current state instead of a plan made before anything happened.
> 
> You're right that the plan happens once, but we're always working off of the cluster's real state. We persist all resource updates back to the state as the update runs, so each resource always has access to the latest view of the database state. Planning each node individually will produce identical results, except it will duplicate some operations we would otherwise run only once.
> 
> > we don't currently block a Spock version change through update database, that's what spockMajorVersionChanged fixes here. But fixing that alone doesn't make update database safe for this, because of the first point. It just means it correctly says no now instead of doing something we can't verify.
> 
> This is a good change, but we should discuss what we should and shouldn't allow. For example, we might want to allow a major version update, but we should probably block downgrades.
> 
> In its current state, I'm not convinced we need a dedicated endpoint and so much new code for this process. I think we can do a much slimmer implementation that just adds the API validation and the replication status validation.

@jason-lynch I agree. After taking another look at the existing implementation, I don't think we need a dedicated workflow for major upgrades.

`RollingUpdateNodes/applyPlans` already performs what we need for a major upgrade—it computes a state diff for each node and applies the plans sequentially (replicas first, then the primary). So the "one node at a time" behavior is already part of the existing update flow.

I think we can simplify this significantly by:

* Relaxing `spockMajorVersionChanged()` in `validate.go` to reject only **downgrades**, matching `FindUpgrade`'s forward-only behavior. That would allow a spec with a higher Spock major version to flow through the normal `update-database` path instead of being blocked.
* Wiring the `VerifyNodeReplicating` activity (already implemented in this PR) into the existing per-plan loop in `applyPlans` (`common.go`), gated on whether the plan changes the node's resolved Spock version. This gives us the fail-fast replication verification without introducing a separate workflow.

With that approach, we could remove:

* `apply-major-upgrade` API action
* `MajorVersionUpgrade` workflow
* Per-node `spock_version` override and the `ApplyMajorUpgradeSpecChange` bypass
* `FindMajorUpgrade`
* `node_order`

The existing `update-database` flow already resolves each node's image from the manifest and already handles node ordering.

There are two behavior changes I'd like to call out explicitly:

1. **Custom node ordering would no longer be supported.** `update-database` doesn't currently expose a mechanism to choose which node is upgraded first. If that's a requirement for major upgrades, this approach won't address it.

2. **The explicit "deliberate action" gate would be removed.** Today, a major upgrade requires calling a dedicated endpoint. Folding it into `update-database` means that any spec update containing a newer Spock major version—including an accidental one—would initiate a one-way, protocol-changing upgrade. The replication verification provides a safety check, but what do you think about this tradeoff?

@moizpgedge

PR #457, opened 2026-08-19
feat: add foundational control plane support for Spock 6 clusters

moizpgedge:
## Summary

This adds the foundational Control Plane support for Spock 6 clusters. While verifying it, I also found and fixed three real bugs that weren't part of the original ticket, including one that was actually failing CI on this PR.

## Changes

- Parses Spock pre-release versions (e.g. `6.0.0-beta.1`) in `ds/versions.go`.
- Adds a Spock 6 dev manifest entry. It's available but never picked as the default.
- Makes `spock.progress` queries use the right column depending on version (`remote_lsn` on 5.x, `remote_commit_lsn` on 6.x).
- Adds `VerifySubscriptionReplicatingResource`, which checks that a newly enabled subscription actually starts replicating, and fails loudly if it doesn't.
- Fixes `EnablePeerSubscriptions`. This was a long standing bug, unrelated to Spock 6, where peer subscriptions could get stuck permanently disabled when adding a node to a database that already had 3 or more nodes.
- Adds `e2e/spock6_add_node_test.go`, an automated test that adds a node to a real Spock 6 cluster, plus a nightly CircleCI job so this keeps getting tested even if nobody touches the branch.
- Removes `ReplicationSlotName` and inlines it, same cleanup as PLAT-717 (this branch had cherry-picked the version from before that cleanup happened).
- Fixes `output_plugin_libraries` not including `spock_output` for the Spock 6 dev image. The check was based on the Postgres version we declared, but the Spock 6 image uses a floating tag that can quietly resolve to a newer Postgres minor than what we declared. Now any Spock 6 build always gets the allow list, regardless of the declared Postgres version.
- Fixes a pre-existing test, `TestCreateDbWithVersions`, that was crashing its entire test run when it hit that same version mismatch. It used `log.Fatalf` instead of `t.Fatalf`, which kills the whole test process instead of just failing the one subtest, so every other test running in parallel got reported as failed too. Switched it to `t.Fatalf` and made the version check only compare the major version for Spock 6 entries, since pinned Spock 5.x checks still need to match exactly.

## Testing

- `go test ./server/internal/ds/... -run TestParseVersion -v`
- `go test ./server/internal/postgres/... -run TestDefaultGUCsOutputPluginLibraries -v`
- `go test -tags e2e_test ./e2e/... -run TestSpock6AddNode -v -timeout 15m -count=1`, ran 5 times in a row cleanly, about 195 to 210 seconds each
- `go test -tags e2e_test ./e2e/... -run 'TestCreateDbWithVersions/postgres_18.4_with_spock_6' -v -timeout 8m -count=1`, confirmed it no longer crashes the suite
- Manually added nodes to a real Spock 6 cluster (went from 2 to 3 to 4 nodes) and confirmed full mesh replication. Also confirmed `output_plugin_libraries` includes `spock_output` and that Spock 6 slots automatically get `failover=t`
- Repeated the same checks against Spock 5.x to confirm nothing regressed: `output_plugin_libraries` stays unset, `failover=f`, and `spock.progress` still has the right columns
- Forced the subscription enablement failure case live to make sure it actually reports an error instead of hanging silently (see notes below)
- `circleci config validate .circleci/config.yml`

## Checklist

- [ ] Documentation updated (if needed)
- [x] Issue is linked (branch name or URL in PR description): [PLAT-718](https://pgedge.atlassian.net/browse/PLAT-718)
- [x] Breaking changes (if any) are clearly called out in the PR description: none

## Notes for Reviewers

While testing the subscription enablement failure case, I found a separate issue worth flagging. There's a resource called `WaitForSyncEventResource` that runs before `VerifySubscriptionReplicatingResource` in the same pipeline. It's pre-existing and not part of this PR. If a subscription stays disabled or down, this resource treats that as temporary and just keeps waiting forever, with no internal timeout, instead of failing with a clear error like `VerifySubscriptionReplicatingResource` does. So if a subscription's apply worker genuinely never starts, the task can hang indefinitely instead of failing loudly. This doesn't block this PR, but it's probably worth its own follow-up ticket.

Also worth knowing: the `output_plugin_libraries` fix always says yes for any Spock 6 build, rather than tracking the exact Postgres minor. That's a deliberate tradeoff since the manifest entry points at a floating tag we can't pin down ahead of time. I confirmed live that setting this on a Postgres minor below the gate is not harmless, Postgres actually rejects it with "unrecognized configuration parameter." If Spock 6 ever ships on a minor below the gate, this will need a second look.

[PLAT-718]: https://pgedge.atlassian.net/browse/PLAT-718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.circleci/config.yml
```diff
@@ -137,6 +137,31 @@ jobs:
           when: always
       - store_test_results:
           path: .
+  test_e2e_spock6:
```
jason-lynch, line 140: Do we need a nightly job for this? It looks like we've only been publishing new images once every few weeks.

server/internal/database/operations/populate_nodes.go
```diff
@@ -85,6 +85,68 @@ func PopulateNodes(existing, new []*NodeResources) (*resource.State, error) {
 	return merged, nil
 }
 
+// EnablePeerSubscriptions returns a diff that enables the peer subscriptions
```
jason-lynch: This comment is inaccurate. Every create and update database process finishes with a desired end state, which we compute in `end.go`. Enabled subscriptions are part of that end state: https://github.com/pgEdge/control-plane/blob/4d64d7f928f97796167ca85557bb2f8a4e1037ea/server/internal/database/operations/end.go#L47-L51

So in an add-node operation, we enable subscriptions near the end. If you look at the golden_test diff in this PR, you'll see that we were already enabling the subscription on line 206. This change just enables the subscription earlier in the flow, which I don't think is necessary.

Could you please explain more about what problem you saw that led you to make this change?

server/internal/orchestrator/swarm/version-manifest.json
```diff
@@ -92,6 +92,12 @@
         "image": "pgedge-postgres:18.4-spock5.0.10-standard-1",
         "stability": "stable",
         "default": true
+      },
+      {
+        "postgres_version": "18.4",
```
jason-lynch: This is out of date. This mutable tag is now on `18.6`, which is probably why you had issues with the output plugins setting.

server/internal/postgres/gucs.go
```diff
@@ -19,10 +19,37 @@ var minOutputPluginLibrariesVersions = map[uint64]*ds.Version{
...
+// rejects it outright with "unrecognized configuration parameter", which
+// would fail startup, not just no-op. If Spock 6 is ever built against
+// such a minor, this unconditional-yes needs revisiting. Spock 5.x
+// behavior is unchanged — it still depends solely on the exact Postgres
+// minor check below, which is exactly why it never hits this failure mode.
 func needsOutputPluginLibraries(version *ds.PgEdgeVersion) bool {
 	if version == nil || version.PostgresVersion == nil {
 		return false
 	}
+	if version.SpockVersion != nil {
```
jason-lynch: As you noted in the comment, the spock version is unrelated to this GUC. Like I mentioned above, I think you hit this because the postgres version in the manifest JSON is outdated. Could you please fix the manifest JSON and try testing without this change? Please let me know if it doesn't work without this change, and we can look at it together.

e2e/spock6_add_node_test.go
```diff
@@ -0,0 +1,101 @@
...
+				Swarm: &controlplane.SwarmOpts{Image: pointerTo(spock6DevImage)},
+			},
+		}
+	}
+
+	t.Log("Step 1: Creating 2-node Spock 6 database fixture")
+	db := fixture.NewDatabaseFixture(ctx, t, &controlplane.CreateDatabaseRequest{
+		Spec: &controlplane.DatabaseSpec{
+			DatabaseName:    dbName,
+			PostgresVersion: pointerTo("18.4"),
```
jason-lynch: This image is at 18.6 now.

server/internal/database/operations/populate_nodes.go
moizpgedge: You're right, my comment was wrong. I turned this phase off and added a node to a live 3 node Spock 6 cluster, and every subscription, including the peer ones, went to replicating on its own, just like you said, driven by end.go. There's no bug here.

I missed that end.go already re enables these later in the same operation, that's what led to the wrong "permanently disabled" comment.

I kept the enable call though, because the verify step right after it needs the subscription to already be enabled to check it. By the time end.go runs, this one's just a no op. Comment's fixed to explain that now. If you'd rather I move the verify step to run after end.go so this whole phase can go away, happy to do that too, just didn't want to touch shared code without checking with you first.

jason-lynch: This check is a good idea, but I think we could implement it in a way that works everywhere. I mentioned in your other PR that it could make sense to raise an error whenever a subscription goes from healthy to unhealthy: https://github.com/pgEdge/control-plane/pull/456#issuecomment-532882156

We could extend that to raise an error whenever a subscription goes from healthy, nonexistent, or disabled to unhealthy, which would apply here too.

Could you please remove that verification step/resource and the new enable subscription step from this PR? We can create a dedicated ticket to add the replication status check, and I'll fill in some implementation suggestions.

jason-lynch (approved):

PR #458, opened 2026-08-20
feat: add native failover-slot support for Spock 6

moizpgedge:
## Summary

This adds the native failover slot support Spock 6 needs on Postgres 17 and up.

## Changes

- Added a check, `postgres.NeedsNativeFailoverSlots`, that only turns any of this on when the database is running Spock 6 or newer and Postgres 17 or newer. Everything else in this PR sits behind that one check.
- `CreateReplicationSlot` now accepts a `failover` flag. When that flag is off, the SQL it produces is exactly what it was before this change, so there is nothing new for clusters that don't qualify.
- Added `NativeFailoverSlotGUCs`, which turns on `sync_replication_slots` as part of the normal Patroni config generation once a database crosses that gate.
- Added a new piece to the instance monitor, `reconcileSynchronizedStandbySlots`, that keeps the `synchronized_standby_slots` setting pointed at whatever physical standby slots actually exist on the current primary. It checks and recomputes this fresh every five seconds instead of remembering a cached value, so if something only half succeeds (say the config gets patched but the reload doesn't happen), it just corrects itself on the next check rather than getting stuck. It also skips temporary slots on purpose, since those disappear the moment the session that made them ends and were never something worth pointing this setting at.

## Testing

- Added unit tests covering the version gating, the exact SQL `CreateReplicationSlot` produces with the flag on and off, and the temporary slot exclusion.
- Ran this against a real Spock 6 cluster and confirmed the slot actually gets created with failover on, both right when a database is first created and again when adding a node later.
- The part I really wanted to get right was the failover itself, so I set up an actual two instance Patroni pair on separate machines and forced a real failover through the API. Did this twice, swapping which side was primary each time, and both times `synchronized_standby_slots` was already correct on the new primary by the time the failover finished, no extra nudge needed. Wrote data before the failover and after it and confirmed it made it to both sides either way.
- Also checked the regression direction directly: created a physical replication slot on a Spock 5.x database on purpose, to see if it would trick the monitor, and confirmed it just left everything alone like it should.

## Checklist

- [ ] Documentation updated (if needed)
- [x] Issue is linked (branch name or URL in PR description): [PLAT-719](https://pgedge.atlassian.net/browse/PLAT-719)
- [x] Breaking changes (if any) are clearly called out in the PR description: none

[PLAT-719]: https://pgedge.atlassian.net/browse/PLAT-719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/internal/monitor/instance_monitor.go
+++ b/server/internal/monitor/instance_monitor.go
@@ -5,12 +5,14 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/rs/zerolog"
 
 	"github.com/pgEdge/control-plane/server/internal/certificates"
 	"github.com/pgEdge/control-plane/server/internal/database"
+	"github.com/pgEdge/control-plane/server/internal/ds"
 	"github.com/pgEdge/control-plane/server/internal/patroni"
 	"github.com/pgEdge/control-plane/server/internal/postgres"
 	"github.com/pgEdge/control-plane/server/internal/utils"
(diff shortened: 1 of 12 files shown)


server/internal/patroni/gucs.go
```diff
@@ -16,6 +16,20 @@ var dynamicGUCs = ds.NewSet(
 	"max_replication_slots",
 	"wal_keep_segments",
 	"wal_keep_size",
+	// Reload-safe, kept identical across every instance in the node via
```
jason-lynch: The values in `dcsGUCs` and `dynamicGUCs` come from here: https://patroni.readthedocs.io/en/latest/patroni_configuration.html#important-rules

`sync_replication_slots` isn't listed on that page, so it doesn't belong in this list, and we can just treat it like any other setting.

server/internal/patroni/gucs.go
```diff
@@ -16,6 +16,20 @@ var dynamicGUCs = ds.NewSet(
...
+	// promote any of them to primary later.
+	"sync_replication_slots",
+	// Never generated as a static default (see NativeFailoverSlotGUCs'
+	// doc comment) -- its correct value depends on live replication
+	// topology, so it's only ever pushed here directly via the Patroni
+	// REST client's PatchDynamicConfig, by InstanceMonitor's runtime
+	// reconciliation (see server/internal/monitor/instance_monitor.go).
+	// Listed here purely so this file stays the one place documenting
+	// every GUC this codebase manages through Patroni's DCS, reload-safe.
+	"synchronized_standby_slots",
```
jason-lynch: This setting is related to synchronous physical replicas. We currently support only asynchronous replicas, so we shouldn't set this parameter.

server/internal/monitor/instance_monitor.go
```diff
@@ -175,6 +188,126 @@ func (m *InstanceMonitor) populateFromDbConn(
...
+		// Logged and swallowed rather than propagated: this is a
+		// best-effort background reconciliation, not a health signal.
+		// Letting it fail the whole status collection here would report
+		// an otherwise-healthy primary as errored (discarding the
+		// version/subscription data this same pass already collected)
+		// over what's usually a transient Patroni REST hiccup -- and
+		// since the reconciliation is self-correcting on every poll (see
+		// its own doc comment), nothing is lost by retrying next tick
+		// instead of surfacing this as an instance-level error now.
+		if err := m.reconcileSynchronizedStandbySlots(ctx, conn, info, pgVersion, spockVersion); err != nil {
```
jason-lynch: Why is this necessary? As I mentioned in another comment, we don't support synchronous replicas right now, so I don't think we should set `synchronized_standby_slots`.

server/internal/postgres/gucs.go
```diff
@@ -67,6 +67,70 @@ func DefaultGUCs(version *ds.PgEdgeVersion) map[string]any {
...
+//
+// synchronized_standby_slots is deliberately NOT set here even when the
+// gate passes: its correct value is the current set of physical standby
+// slot names, which isn't knowable at config-generation/bootstrap time
+// (no instances exist yet, let alone standbys) and changes over the
+// node's lifetime as replicas are added/removed or a failover promotes a
+// different primary. That value is instead computed from live replication
+// state and kept in sync at runtime — see
+// server/internal/monitor/instance_monitor.go.
+func NativeFailoverSlotGUCs(version *ds.PgEdgeVersion) map[string]any {
```
jason-lynch: `DefaultGUCs` in this file already takes the pgEdge version and has similar logic for `output_plugin_libraries`. Is there a reason to add this new function rather than just incorporating this logic into `DefaultGUCs`?

server/internal/patroni/gucs.go
moizpgedge: Checked the actual docs and it turns out this isn't tied to synchronous replication at all, it works the same with async standbys. But there is a real catch worth knowing about. If a physical standby lags or drops, the primary just waits for it before sending changes to Spock peers, no timeout. And since a node can already span multiple hosts today, that's not a hypothetical, a slow or disconnected replica could stall a node's whole outbound replication.

So it's not exactly the sync replication issue you mentioned, but there is a real tradeoff here worth talking about before we lock t

server/internal/monitor/instance_monitor.go
moizpgedge: Same answer as the other thread, this is what makes a failover slot safe to use after a promotion. Without it, a promoted standby has no guarantee it got the WAL a failover enabled slot needed.

One thing worth flagging though, if a physical standby lags or drops, the primary waits for it with no timeout, and nodes can already span multiple hosts today, so that's a real risk. Mechanism's doing what it's supposed to, just introduces a new way a slow replica could stall replication. Happy to keep it or talk through bounding it if that's a concern.

jason-lynch: Thank you for the clarification. I was confused because the wording in the Postgres docs is very similar to synchronous replication, but I see the distinction now. You're right that this setting carries a risk, similar to the risk with synchronous replicas. But I also see that it's listed in the Spock documentation, and they're the experts on this.

That said, I'd like to try to set this parameter in a more standard way and not overload this instance monitor. I don't think this needs to be dynamic, because topology changes (adding or removing replicas) will always flow through our database update process. Additionally, from some quick testing, it looks like:

- Patroni's physical replication slots have predictable names
  - Their slot name function is here: https://github.com/patroni/patroni/blob/1150a18cbd66303ecbea439ead8e2a5c221883f1/patroni/dcs/__init__.py#L36-L55
  - Our member names are just the instance ID, so for example, instance `storefront-n1-9ptayhma` would map to slot name `storefront_n1_9ptayhma`
  - It's safe to consider this naming scheme as stable. Their implementation has been stable for at least 8 years, and modifying it would be a breaking change.
- `synchronized_standby_slots` is ignored on replicas

Given all that, we can incorporate this into our normal configuration process. I suggest adding a `PeerInstanceIDs string[]` property to `database.InstanceSpec` that gets populated in `database.Spec.NodeInstances()` (make sure it's included in `database.InstanceSpec.Clone()` as well). Then you can propagate that setting into `common.PatroniConfigGenerator`. Then you can compute `synchronized_standby_slots` from those peer instance IDs. You have options for where that computation happens. It could make sense to do it in `DefaultGUCs` since it's already deciding on `sync_replication_slots`, but regardless, I think that's the right general direction.

server/internal/patroni/gucs.go
jason-lynch: Thank you for the clarification. My other comment still applies here, though. This setting does not belong in this list. Setting this in the DCS means that it will be set to the same value on every instance in the node, which is _not_ what we want. For example, say we have a node `n1` with two instances: `storefront-n1-689qacsi` and `storefront-n1-9ptayhma`. When `storefront-n1-689qacsi` is the primary, it will have a physical replication slot called `storefront_n1_9ptayhma`. Likewise, when `storefront-n1-9ptayhma` is the primary, it will have a replication slot called `storefront_n1_689qacsi`. So, we want to configure this setting as follows:

- On `storefront-n1-689qacsi`, `synchronized_standby_slots` should be `storefront_n1_9ptayhma`
- On `storefront-n1-9ptayhma`, `synchronized_standby_slots` should be `storefront_n1_689qacsi`

I left a suggestion in our other conversation above about how to think about implementing this.

server/internal/postgres/create_db.go
```diff
@@ -408,6 +423,42 @@ func ReplicationSlotExists(databaseName, providerNode, subscriberNode string) Qu
 	}
 }
 
+// PhysicalReplicationSlotNames lists every permanent (non-temporary)
```
jason-lynch: I think you'll remove this function after making the other changes I suggested, but I want to point out that we should avoid these types of comments. This comment is very long and describes code and behavior that aren't adjacent to it. If we decide to use this function for a different purpose, or if we change the behavior it describes, the comment would mislead future readers. Also, in this case, the function is well-named and easy to read and understand, so I wouldn't add a comment to this type of function unless this were a publicly consumed library.

As a rule of thumb, I think it's best to avoid long, multi-paragraph comments unless the code it describes is very tricky. Otherwise, they add a maintenance burden and potential issues that a compiler can't catch. If you need to describe behavior once to a reviewer, you can always leave comments on your PR after you open it.

Could you please double-check the other long comments in this PR and make sure they're really necessary?

jason-lynch (approved):

PR #461, opened 2026-09-01
fix: instance health reflects in db state

GabriellePoncey:
## Summary
Updated storedToDtabase to inspect the database's instance and update the database to degraded if instance states indicate this is appropriate. Delete database modified to check "from" state via the stored state rather than etcd to prevent mismatch.

## Changes 

- storedToDatabase  downgrades an available database to degraded if any of its instances is degraded, failed, unknown, or stopped.
- database_test.go covers and tests this logic
- DeleteDatabase now uses GetStoredDatabaseState (raw stored value) instead of db.State (computed) for its state-transition guard
  
PLAT-333

### DIFF
--- a/server/internal/api/apiv1/post_init_handlers.go
+++ b/server/internal/api/apiv1/post_init_handlers.go
@@ -475,7 +475,11 @@ func (s *PostInitHandlers) DeleteDatabase(ctx context.Context, req *api.DeleteDa
 		return nil, ErrDatabaseNotModifiable
 	}
 
-	prevState := db.State
+	// Use the raw stored state as the guard for this transition.
+	prevState, err := s.dbSvc.GetStoredDatabaseState(ctx, db.DatabaseID)
+	if err != nil {
+		return nil, apiErr(err)
+	}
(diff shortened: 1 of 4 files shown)


server/internal/api/apiv1/post_init_handlers.go
```diff
@@ -475,7 +475,11 @@ func (s *PostInitHandlers) DeleteDatabase(ctx context.Context, req *api.DeleteDa
 		return nil, ErrDatabaseNotModifiable
 	}
 
-	prevState := db.State
+	// Use the raw stored state as the guard for this transition.
+	prevState, err := s.dbSvc.GetStoredDatabaseState(ctx, db.DatabaseID)
```
tsivaprasad, line 479: Looks good.
Minor: `DeleteDatabase` now calls `GetStoredDatabaseState` (`post_init_handlers.go:479`), which re-reads the same `StoredDatabase` record already fetched by `GetDatabase` just above. We could avoid the extra database round trip by carrying the raw stored state on `Database` (for example, a `RawState` field populated in `storedToDatabase`) and using `db.RawState` here instead. Not blocking.

tsivaprasad (approved):
Looks good.
Thanks

PR #462, opened 2026-09-02
docs: document Spock 6 preview support

tsivaprasad:
## Summary
This PR documents the preview support for Spock 6, native replication slot failover behavior, and the corresponding changelog entries.

## Changes
- Add a "Spock 6 Preview Images" section to `docs/using/image-management.md`
  explaining how to use the Spock 6 preview manifest entry (`"stability":
  "dev"`), either by setting `postgres_version`/`spock_version` directly
  or by pinning `orchestrator_opts.swarm.image`, and its exclusion from
  the image-upgrade action and `available_upgrades`.
- Add a note to `docs/prerequisites/concepts.md`'s Instances section
  describing automatic native replication slot failover
  (`sync_replication_slots`/`synchronized_standby_slots`) for databases
  running Spock 6 on Postgres 17+.
- Add a changelog entry for Spock 6 foundational support.

## Testing
- Docs-only change; no code paths touched.

Verification:
```
curl -X POST http://localhost:3000/v1/databases \
        -H 'Content-Type:application/json' \
        --data '{
            "id": "example",
            "spec": {
                "database_name": "example",
                "database_users": [
                    {
                        "username": "admin",
                        "db_owner": true,
                    }
                ],
                "postgres_version": "18.6",
                "spock_version": "6",
                "orchestrator_opts": {
                    "swarm": {
                        "image": "ghcr.io/pgedge/pgedge-postgres:18-spock6-standard"
                    }
                },
                "nodes": [
                ]
            }
        }'

```
```
cp1-req list-databases
HTTP/1.1 200 OK
Content-Length: 943
Content-Type: application/json
Date: Wed, 02 Sep 2026 05:36:29 GMT

{
  databases: [
    {
      created_at: "2026-09-02T05:35:14Z"
      id: "example"
      instances: [
        {
          created_at: "2026-09-02T05:35:17Z"
          host_id: "host-1"
          id: "example-n1-689qacsi"
          node_name: "n1"
          postgres: {
            patroni_state: "running"
            role: "primary"
            version: "18.6"
          }
          spock: {
            read_only: "off"
            subscriptions: [
              {
                name: "sub_n2_n1"
                provider_node: "n2"
                status: "replicating"
              }
            ]
            version: "6.0.0"
          }
          state: "available"
          status_updated_at: "2026-09-02T05:36:26Z"
          updated_at: "2026-09-02T05:35:50Z"
        }
        {
          created_at: "2026-09-02T05:35:17Z"
          host_id: "host-2"
          id: "example-n2-9ptayhma"
          node_name: "n2"
          postgres: {
            patroni_state: "running"
            role: "primary"
            version: "18.6"
          }
          spock: {
            read_only: "off"
            subscriptions: [
              {
                name: "sub_n1_n2"
                provider_node: "n1"
                status: "replicating"
              }
            ]
            version: "6.0.0"
          }
          state: "available"
          status_updated_at: "2026-09-02T05:36:26Z"
          updated_at: "2026-09-02T05:35:50Z"
        }
      ]
      state: "available"
      updated_at: "2026-09-02T05:35:14Z"
    }
  ]
}
```
## Checklist

- [x] Documentation updated (if needed)

[PLAT-734](https://pgedge.atlassian.net/browse/PLAT-734)

[PLAT-734]: https://pgedge.atlassian.net/browse/PLAT-734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/unreleased/Fixed-20260821-141322.yaml
+++ b/changes/unreleased/Fixed-20260821-141322.yaml
@@ -1,3 +1,0 @@
-kind: Fixed
-body: Fixed peer subscriptions remaining permanently disabled when adding a node to a database that already had three or more nodes.
-time: 2026-08-21T14:13:22.000000+00:00
--- a/changes/unreleased/Fixed-20260821-141322.yaml
+++ b/changes/unreleased/Fixed-20260821-141322.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: Fixed peer subscriptions remaining permanently disabled when adding a node to a database that already had three or more nodes.
+time: 2026-08-21T14:13:22.000000+00:00
--- a/docs/using/image-management.md
+++ b/docs/using/image-management.md
@@ -266,8 +266,8 @@ value in `orchestrator_opts.swarm.image` when creating the database:
 
 ### Spock 6 Preview Images
 
-Spock 6 is available today as a preview manifest entry (`"stability":
-"dev"` in the version manifest), currently paired with Postgres 18.6.
+Spock 6 is available as a preview manifest entry (`"stability": "dev"`
+in the version manifest), currently paired with Postgres 18.6.
 Preview entries are excluded from image upgrades (see below) and are
 never chosen for a database that omits `postgres_version` and
 `spock_version`, but you can still request one directly by setting
--- a/docs/using/image-management.md
+++ b/docs/using/image-management.md
@@ -267,15 +267,17 @@ value in `orchestrator_opts.swarm.image` when creating the database:
 ### Spock 6 Preview Images
 
 Spock 6 is available as a preview manifest entry (`"stability": "dev"`
-in the version manifest), currently paired with Postgres 18.6.
+in the version manifest), currently paired with Postgres 18.6. Spock 6
+itself supports Postgres 15 through 19; the version manifest currently
+offers only this one Postgres 18.6 pairing as a preview. This preview
+is available for Docker Swarm deployments only; the systemd
+orchestrator does not currently support Spock 6 packages.
+
 Preview entries are excluded from image upgrades (see below) and are
 never chosen for a database that omits `postgres_version` and
-`spock_version`, but you can still request one directly by setting
-those fields to a matching manifest entry, or by pinning the image
-explicitly with `orchestrator_opts.swarm.image`.
-
-The following request creates a database pinned to the Spock 6 preview
-image:
+`spock_version`. To create a new database on the Spock 6 preview
+image, set `postgres_version` and `spock_version` to match the
+manifest entry:
 
 === "curl"
 
@@ -295,11 +297,6 @@ image:
                 ],
                 "postgres_version": "18.6",
                 "spock_version": "6",
-                "orchestrator_opts": {
-                    "swarm": {
-                        "image": "ghcr.io/pgedge/pgedge-postgres:18-spock6-standard"
-                    }
-                },
                 "nodes": [
                     { "name": "n1", "host_ids": ["host-1"] },
                     { "name": "n2", "host_ids": ["host-2"] }
@@ -308,11 +305,25 @@ image:
         }'
     ```
 
+Overriding `orchestrator_opts.swarm.image` is only necessary if you need
+an image other than the manifest default for that version pair, such as
+pinning a specific build. See
+[Using a Custom Image](#using-a-custom-image).
+
 !!! note
 
     The Spock 6 preview image tracks the latest Spock 6 build against
     Postgres 18, so its resolved Postgres minor version can advance ahead
     of the manifest entry's declared `postgres_version`.
+
+!!! warning
+
+    This preview supports only creating a new database on Spock 6. It
+    does not support upgrading an existing database from Spock 5.x to
+    Spock 6. The Control Plane does not validate or block a
+    `spock_version` change on an existing database's spec, but doing so
+    is unsupported and can break replication, since a Spock 6
+    subscription cannot sync from a Spock 5.x peer.
 
 !!! warning
 
--- a/docs/using/image-management.md
+++ b/docs/using/image-management.md
@@ -264,6 +264,63 @@ value in `orchestrator_opts.swarm.image` when creating the database:
     Digest pinning guarantees that you run a specific immutable image even
     if the tag is later reassigned to a different image in the registry.
 
+### Spock 6 Preview Images
+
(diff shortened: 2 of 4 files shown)


docs/using/image-management.md
```diff
@@ -264,6 +264,63 @@ value in `orchestrator_opts.swarm.image` when creating the database:
...
+                        "attributes": ["SUPERUSER", "LOGIN"]
+                    }
+                ],
+                "postgres_version": "18.6",
+                "spock_version": "6",
+                "orchestrator_opts": {
+                    "swarm": {
+                        "image": "ghcr.io/pgedge/pgedge-postgres:18-spock6-standard"
+                    }
+                },
```
jason-lynch: Overriding the image isn't necessary unless you want a different image than what's in our manifest. Most users can just set `spock_version: 6` in the spec.

changes/unreleased/Fixed-20260821-141322.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Fixed
```
jason-lynch: Could you please remove this file? This issue was a misunderstanding, and Moiz removed this change before we merged his PR. Subscriptions are always re-enabled at the end of the update workflow.

jason-lynch (commented):
I have a few questions:

- Is this preview support only for Docker Swarm, or should we also support systemd? 
  - Right now, the systemd orchestrator filters out Spock 6 packages.
  - If we do want to support systemd, we should also make sure the documentation is clear that this preview does not support upgrading an existing database to Spock 6. You have a note about it in the context of images, but our validation is on the `spock_version` field, not just the image.
- Should we be using the mutable image tag in the manifest?
  - Our update workflow depends on diffs between the current and desired state, so I'm worried that might make it difficult for users to update their database.
  - We could just use the normal immutable tags and then keep the manifest.json up-to-date as new images are built and tested.
- Should we also add the Postgres 16 and 17 + Spock 6 images to our manifest?
  - We currently only have one for Postgres 18.6

tsivaprasad:
> I have a few questions:
> 
> * Is this preview support only for Docker Swarm, or should we also support systemd?
>   
>   * Right now, the systemd orchestrator filters out Spock 6 packages.
>   * If we do want to support systemd, we should also make sure the documentation is clear that this preview does not support upgrading an existing database to Spock 6. You have a note about it in the context of images, but our validation is on the `spock_version` field, not just the image.
> * Should we be using the mutable image tag in the manifest?
>   
>   * Our update workflow depends on diffs between the current and desired state, so I'm worried that might make it difficult for users to update their database.
>   * We could just use the normal immutable tags and then keep the manifest.json up-to-date as new images are built and tested.
> * Should we also add the Postgres 16 and 17 + Spock 6 images to our manifest?
>   
>   * We currently only have one for Postgres 18.6

Thanks for the review. I checked each of these points:

- **Docker Swarm vs. systemd:** This preview is currently Docker Swarm only. The systemd orchestrator has supportedSpockVersions hardcoded to ["50"], so it cannot discover or use Spock 6 packages today. I’ve updated the documentation to make the Docker Swarm-only scope explicit.

- Existing database upgrades: I’ve also clarified that Spock 6 preview support is create-only. There is currently no supported upgrade path from an existing Spock 5.x database to Spock 6. One additional detail I found is that ValidateChangedSpec does not actually prevent changing spock_version on an existing database; it only validates changes to the PostgreSQL major version. So a spock_version bump is currently unsupported/unsafe rather than explicitly blocked by validation.

- **Mutable vs. immutable image tags**: I agree this should be considered together with the update workflow. Since updates depend on detecting differences between the current and desired state, immutable tags are likely safer and make image changes explicit. Rather than changing the tagging strategy as part of this documentation PR, I suggest settling this when we update the manifest with the additional Spock 6 images.

- **Postgres 16/17 + Spock 6**: I checked the upstream registry, and Spock 6 images for PostgreSQL 16 and 17 already exist. So this is not an image-build availability gap; they are simply not present in our manifest yet. I suggest validating PostgreSQL 16/17 + Spock 6 support and adding those entries to manifest.json in a separate PR. We can resolve the mutable-vs.-immutable tag question there as well, and I’ll follow up with the corresponding documentation update once that lands.

For now, I’ll keep this PR scoped to documenting the existing PostgreSQL 18.6 + Spock 6 Docker Swarm preview. 
What do you think?

jason-lynch:
> Docker Swarm vs. systemd: This preview is currently Docker Swarm only. The systemd orchestrator has supportedSpockVersions hardcoded to ["50"], so it cannot discover or use Spock 6 packages today. I’ve updated the documentation to make the Docker Swarm-only scope explicit.

Thanks for making this change!

> Existing database upgrades: I’ve also clarified that Spock 6 preview support is create-only. There is currently no supported upgrade path from an existing Spock 5.x database to Spock 6. One additional detail I found is that ValidateChangedSpec does not actually prevent changing spock_version on an existing database; it only validates changes to the PostgreSQL major version. So a spock_version bump is currently unsupported/unsafe rather than explicitly blocked by validation.

Thank you for pointing this out! I remembered reviewing this change from Moiz, but I forgot that we didn't merge it. This is a blocker for the release. Our validation should disallow changes to the Spock major version for existing nodes and databases. Could you please fix this either in this PR or another one?

>  - Mutable vs. immutable image tags: I agree this should be considered together with the update workflow. Since updates depend on detecting differences between the current and desired state, immutable tags are likely safer and make image changes explicit. Rather than changing the tagging strategy as part of this documentation PR, I suggest settling this when we update the manifest with the additional Spock 6 images.
>  - Postgres 16/17 + Spock 6: I checked the upstream registry, and Spock 6 images for PostgreSQL 16 and 17 already exist. So this is not an image-build availability gap; they are simply not present in our manifest yet. I suggest validating PostgreSQL 16/17 + Spock 6 support and adding those entries to manifest.json in a separate PR. We can resolve the mutable-vs.-immutable tag question there as well, and I’ll follow up with the corresponding documentation update once that lands.
>
> For now, I’ll keep this PR scoped to documenting the existing PostgreSQL 18.6 + Spock 6 Docker Swarm preview.
What do you think?

I'm fine with limiting the preview to 18.6 and Spock 6, but I think the mutable tags are a blocker for this release. I'm raising it now because it will change some of the wording that you're adding in this PR, but if you'd rather edit it again in another PR, then that's OK with me.

jason-lynch (approved):

PR #464, opened 2026-09-08
fix: reject Spock major version change on update

GabriellePoncey:
## Summary

Major-version guard now rejects Spock major version changes on an existing database. 

## Changes

- ValidateChangedSpec rejects Spock major version changes on an existing database
- image-management.md updated to reflect these change

### DIFF
--- a/server/internal/database/service.go
+++ b/server/internal/database/service.go
@@ -1098,7 +1098,7 @@ func majorVersionChanged(old, new *ds.PgEdgeVersion) error {
 	}
 	if oldSpockMajor != newSpockMajor {
 		return fmt.Errorf(
-			"spock major version changed from %d to %d: upgrading the spock major version on an existing database is not supported",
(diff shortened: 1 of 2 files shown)


server/internal/database/service.go
tsivaprasad, line 1051: The Spock major-version validation currently runs only when an instance ID matches between the current and updated specs. Since the instance ID is derived from `hash(hostID, nodeName)`, changing a node's `host_ids` together with `spock_version` generates a new instance ID. As a result, no matching instance is found, the validation is skipped, and the Spock major-version change can go through unvalidated.

Since `spock_version` is database-wide and does not support per-node overrides, we should validate it once at the spec level by comparing `current.SpockVersion` with `updated.SpockVersion`, rather than validating it per instance. This would also prevent duplicate validation errors for multi-node updates.

We should also add tests covering the intended behavior:

* Reject a Spock major-version change.
* Allow a Spock minor-version change.

tsivaprasad (approved):

